PR #59, opened 2016-04-03
Fixed paging problem within SimpleEntityRetriever

### LINKED ISSUE #31: Solving "Virtuoso 22023 Error SR353: Sorted TOP clause specifies more then 11000 rows to sort." error
Hi,

In order to solve this Virtuoso error which happen pretty often as Virtuoso by default fix this number to 10000, changing the way how Silk creates the queries could be a solution. For example changing :

```
SELECT ?a ?v0
FROM <http://dbpedia.org>
WHERE
  {
    ?a a <http://dbpedia.org/ontology/Country> .
    OPTIONAL
    {
        ?a <http://www.w3.org/2000/01/rdf-schema#label> ?v0 .
    }
  }
OFFSET  10000
LIMIT   1000
```

To

```
SELECT ?a ?v0
WHERE
  {
    {
      SELECT ?a ?v0
      FROM <http://dbpedia.org>
      WHERE
        {
          ?a a <http://dbpedia.org/ontology/Country> .
          OPTIONAL
          {
            ?a <http://www.w3.org/2000/01/rdf-schema#label> ?v0 .
          }
        } 
    }
  }
OFFSET 10000
LIMIT 1000
```

No ?

propi:
In SimpleEntityRetriever the original sparql SELECT statement was wrapped by
another SELECT query to solve a paging problem with ORDER BY clause.
The problem occurs in a Virtuoso remote endpoint. This commit fixes #31

### DIFF
--- a/silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/sparql/SimpleEntityRetriever.scala
+++ b/silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/sparql/SimpleEntityRetriever.scala
@@ -50,10 +50,11 @@ class SimpleEntityRetriever(endpoint: SparqlEndpoint, pageSize: Int = 1000, grap
     val sparqlEntitySchema = SparqlEntitySchema.fromSchema(entitySchema)
     //Select
     var sparql = "SELECT DISTINCT "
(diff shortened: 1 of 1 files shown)


robertisele:
I have added a 'strategy' parameter to the SparqlDataset. The subQuery stragegy uses the pattern from this pull request

PR #116, opened 2017-06-21
added example view widget and api call

JorgeMartinezPizarro:
- create new component ExampleView
- added (whenever an id is available) to both value and object rule
- fix event propagation minor warning
- implemented the api call according to [DI Dokumentation](https://docs.eccenca.com/docs/develop/eccenca-DataIntegration/v3.4.0-160-g5e5773fe-dirty/eccenca-DataIntegration_v3.4.0-160-g5e5773fe-dirty_css-rel_DeveloperManual.html#transformtasksprojecttransformationtaskpeakrule)


JorgeMartinezPizarro:
Maybe we should wait until the styling is finished

PR #123, opened 2017-06-27
Vocabulary Info

leipert:


leipert:
- @haschek How to handle empty description? Simply hide the info element?
- @haschek How to handle loooooong descriptions?

PR #129, opened 2017-07-05
Hierarchical Mapping, minor fixes.

JorgeMartinezPizarro:
It fixes CMEM-471, CMEM-469 and CMEM-474


JorgeMartinezPizarro:
It fixes CMEM-471, CMEM-469 and CMEM-474

PR #139, opened 2017-08-28
Fix/remove double auto complete entries

KayVollers:


silk-react-components/src/HierarchicalMapping/Components/MappingRule/Forms/AutoComplete.jsx
```diff
@@ -42,19 +42,27 @@ const AutoComplete = React.createClass({
             );
         }
 
+        // only show value entry if it is not same as label
+        let optionValue = false;
```
JorgeMartinezPizarro: Mutable variables are dangerous in case the value changes in another place. Why not use a const with conditional value, ie. const optionValue = (value !==label) ? a : b.

PR #141, opened 2017-09-05
example view available in formulars too

JorgeMartinezPizarro:


haschek:
@JorgeMartinezPizarro It's currently not possible to merge it because of Javascript errors, we need to fix that first.

JorgeMartinezPizarro:
Fixed and added rule validation before showing exampleView in formulars within the last 2 commits.

PR #146, opened 2017-09-19
show example data in rule form requires only sourcePath to be defined

JorgeMartinezPizarro:
example data should be visible even if the rule is not completely defined, it is enough to have parentID and sourcePath


JorgeMartinezPizarro:
I think the endpoint peak/childRule should handle the undefined 'uri' parameter, not the component. Reason for that is that the respond depends only on soucePath, not on the complete rule.

PR #189, opened 2018-08-09
Feature/entity metadata

andreas-schultz:


silk-core/src/main/scala/org/silkframework/entity/metadata/EntityMetadataXml.scala
```diff
@@ -0,0 +1,94 @@
...
+import org.silkframework.failures.FailureClass
+import org.silkframework.runtime.serialization.{ReadContext, SerializationFormat, WriteContext, XmlFormat}
+
+import scala.xml.Node
+
+
+case class EntityMetadataXml(override val metadata: Map[String, LazyMetadata[_, Node]] = Map[String, LazyMetadata[_, Node]]()) extends EntityMetadata[Node] {
+
+  def this(rawMetadata: String) = {
+    this(XmlSerializer.fromString(rawMetadata, XmlFormat.MIME_TYPE_TEXT)(ReadContext()))
```
robertisele: Couldn't the parsing of the metadata here be lazy as well?

PR #250, opened 2019-03-04
Feature/copy and clone mapping rules

amartinezpizarro:
CMEM-1818


silk-react-components/src/HierarchicalMapping/Components/MappingRule/ObjectMappingRule.jsx
```diff
@@ -284,6 +291,28 @@ const ObjectRule = React.createClass({
         let entityRelation = false;
         let deleteButton = false;
 
+        const copyButton = isCopiableRule(this.props.type) ? (
```
yeremyantshrach: this can be replaced with `const copyButton = isCopiableRule(this.props.type) && <Button
                className="ecc-silk-mapping__rulesviewer__actionrow-copy"
                raised
                onClick={this.handleCopy}>
                Copy
            </Button>`

silk-react-components/src/HierarchicalMapping/Components/MappingRule/ObjectMappingRule.jsx
```diff
@@ -284,6 +291,28 @@ const ObjectRule = React.createClass({
...
+                className="ecc-silk-mapping__rulesviewer__actionrow-copy"
+                raised
+                onClick={this.handleCopy}>
+                Copy
+            </Button>
+        ) : (
+            false
+        );
+
+        const cloneButton = isCopiableRule(this.props.type) ? (
```
yeremyantshrach: the same is here check the comment above

silk-workbench/silk-workbench-rules/app/controllers/transform/TransformTaskApi.scala
```diff
@@ -144,19 +144,125 @@ class TransformTaskApi extends Controller {
     }
   }
 
-  def appendRule(projectName: String, taskName: String, ruleName: String): Action[AnyContent] = RequestUserContextAction { implicit request => implicit userContext =>
+  def appendRule(projectName: String,
```
nucknorris: We should provide method comments for public methods here.

silk-workbench/silk-workbench-rules/app/controllers/transform/TransformTaskApi.scala
```diff
@@ -144,19 +144,125 @@ class TransformTaskApi extends Controller {
...
+  private def convertRootMappingRule(rule: TransformRule): TransformRule = {
+    rule match {
+      case RootMappingRule(rules, id, metaData) =>
+        ObjectMapping(id, rules = rules, metaData = metaData)
+      case other: TransformRule =>
+        other
+    }
+  }
+
+  def copyRule(projectName: String,
```
nucknorris: We should provide method comments for public methods.

silk-react-components/src/HierarchicalMapping/Components/MappingRule/MappingRule.jsx
```diff
@@ -47,8 +48,12 @@ const MappingRule = React.createClass({
 
     // initilize state
     getInitialState() {
+        const isPasted = sessionStorage.getItem('pastedId') &&
+            (sessionStorage.getItem('pastedId') === this.props.id) &&
+            !sessionStorage.removeItem('pastedId');
```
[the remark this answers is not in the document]
amartinezpizarro: Sure. But I use the same logic to read if it exists, if it's what I want, and if those two are True to remove if from the Storage in less code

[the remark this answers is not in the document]
yeremyantshrach: Yes I understand you but you are passing down to other components 2 type of values

1. `null`
2. `boolean`

It's not good practice that because I wrote that comment, also executing one function multiple time is not a goo practice

PR #257, opened 2019-03-20
fix display text in Mapping editor to always show the target property…

amartinezpizarro:
… and also the label if it exists

CMEM-1628


silk-react-components/test/HierarchicalMapping/Components/MappingRule/MappingRule.test.jsx
```diff
@@ -0,0 +1,139 @@
+import React from 'react';
+import {expect, assert} from 'chai';
```
yeremyantshrach: `import chai, {expect, assert} from 'chai'` instead of importing separately

silk-react-components/test/HierarchicalMapping/Components/MappingRule/MappingRule.test.jsx
```diff
@@ -0,0 +1,139 @@
...
+    uri : "fibo-be-lex-nam:hasName",
+    uriRelative : "name",
+    label : "myLabel",
+    uriLabel : "Has Name",
+    uriLabelRelative : "Name",
+};
+
+const mountMappingRule = (uri, label) => mount(
+    <DragDropContext
+        onDragStart={() => {}}
```
yeremyantshrach: you can use `sinon.spy()` for mocking those functions

PR #271, opened 2019-05-10
Feature/execution report react

andreas-schultz:


silk-react-components/src/HierarchicalMapping/HierarchicalMapping.scss
```diff
@@ -285,6 +286,20 @@ ul.ecc-hierarchical-mapping-error-list {
...
 }
 
+.ecc-silk-mapping__ruleitem-icon-green {
+    font-size: 1.39em;
+    vertical-align: bottom;
+    margin-right: 0.1em;
+    color: to_color($palette-green-800);
+}
+
+.ecc-silk-mapping__ruleitem-icon-red {
```
KayVollers: `.ecc-silk-mapping__ruleitem-icon-green` and `.ecc-silk-mapping__ruleitem-icon-red`can be merged together as most of the props are equal (and always will be)
```suggestion
.ecc-silk-mapping__ruleitem-icon {
    font-size: 1.39em;
    vertical-align: bottom;
    margin-right: 0.1em;
    
    &-red {
        color: to_color($palette-red-800);
    }
    &-green {
        color: to_color($palette-green-800);
    }
```

Pls check before that `.ecc-silk-mapping__ruleitem-icon` not already exists

silk-react-components/src/api/silkRestApi.js
```diff
@@ -0,0 +1,135 @@
+import superagent from '@eccenca/superagent';
+import Promise from 'bluebird';
```
KayVollers: I know the file is only renamed more or less, so all the adds are copy&paste, but is it really intended to user superagent AND bluebird? Both do exactly the same in its core. Fun fact: In future superagent will be replaced by axios (the third big player for http requests)

PR #298, opened 2019-09-05
SPARQL query timeout

andreas-schultz:
limit execution of SPARQL queries in:
- SPARQL endpoint dataset
- SPARQL Select operator

### DIFF
--- a/silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/datasets/SparqlDataset.scala
+++ b/silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/datasets/SparqlDataset.scala
@@ -39,7 +39,12 @@ case class SparqlDataset(
   useOrderBy: Boolean = true,
   @Param(label = "Clear graph before workflow execution",
     value = "If set to true this will clear the specified graph before executing a workflow that writes to it.")
-  clearGraphBeforeExecution: Boolean = true) extends RdfDataset with TripleSinkDataset {
+  clearGraphBeforeExecution: Boolean = true,
+  @Param(
+    label = "SPARQL query timeout (ms)",
+    value = "SPARQL query timeout (select/update) in milliseconds. A value of zero means that the configured default timeout is used." +
+        " If a value greater zero is specified, it overwrites the default timeouts, e.g. configured via silk.remoteSparqlEndpoint.defaults.read.timeout.ms.")
+  sparqlTimeout: Int = 0) extends RdfDataset with TripleSinkDataset {
 
   private val params =
     SparqlParams(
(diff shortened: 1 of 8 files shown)


silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/datasets/SparqlDataset.scala
```diff
@@ -39,7 +39,12 @@ case class SparqlDataset(
   useOrderBy: Boolean = true,
   @Param(label = "Clear graph before workflow execution",
     value = "If set to true this will clear the specified graph before executing a workflow that writes to it.")
-  clearGraphBeforeExecution: Boolean = false) extends RdfDataset with TripleSinkDataset {
+  clearGraphBeforeExecution: Boolean = false,
+  @Param(
+    label = "SPARQL query timeout (ms)",
+    value = "SPARQL query timeout (select/update) in milliseconds. A value of zero means that the configured default timeout is used." +
+        " If a value greater zero is specified, it overwrites the default timeouts, e.g. configured via silk.remoteSparqlEndpoint.defaults.read.timeout.ms.")
```
nucknorris: That sentence is a bit confusing regarding what overwrites what. Proposal:

A value of zero means that the timeout configured via property is used (e.g. configured via silk.remoteSparqlEndpoint.defaults.read.timeout.ms).  To overwrite the configured value specify a value greater than zero.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/tasks/SparqlSelectCustomTask.scala
```diff
@@ -27,7 +27,13 @@ case class SparqlSelectCustomTask(@Param(label = "Select query", value = "A SPAR
                                   limit: String = "",
                                  @Param(label = "Optional SPARQL dataset",
                                    value = "An optional SPARQL dataset that can be used for example data, so e.g. the transformation editor shows mapping examples.")
-                                  optionalInputDataset: SparqlEndpointDatasetParameter = SparqlEndpointDatasetParameter("")) extends CustomTask {
+                                  optionalInputDataset: SparqlEndpointDatasetParameter = SparqlEndpointDatasetParameter(""),
+                                  @Param(
+                                    label = "SPARQL query timeout (ms)",
+                                    value = "SPARQL query timeout (select/update) in milliseconds. A value of zero means that there is no timeout set explicitly." +
+                                        " If a value greater zero is specified this overwrites possible default timeouts.")
```
nucknorris: We should use the same doc here as in SparqlDataset.scala.

andreas-schultz: Since the SPARQL select operator is more generic, we cannot use the same description here. It just passes the timeout down to the input SPARQL endpoint, but the implementation details like default timeouts etc. depend on the SPARQL endpoint implementation and config alone.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/datasets/SparqlDataset.scala
andreas-schultz: Ok, I've updated the description.

PR #340, opened 2019-12-12
Add Knowledge Graph File Upload operator and executor

andreas-schultz:


silk-core/src/main/scala/org/silkframework/dataset/rdf/GraphStoreFileUploadTable.scala
```diff
@@ -0,0 +1,19 @@
...
+trait GraphStoreFileUploadTable {
+  def files: Traversable[FileResource]
+}
+
+case class LocalGraphStoreFileUploadTable(files: Traversable[FileResource], task: Task[TaskSpec]) extends LocalEntities with GraphStoreFileUploadTable {
+  override def entitySchema: EntitySchema = EntitySchema.empty
+
+  override def entities: Traversable[Entity] = Traversable.empty
+
+  override def updateEntities(newEntities: Traversable[Entity]): LocalEntities = EmptyEntityTable(task)
```
chriswartner: @andreas-schultz Here was a small merge issue you may want to take a look at: the update method.

andreas-schultz: It should just return itself. Updating entities has no effect since the entities are always empty for this table.

PR #411, opened 2020-12-01
Execution Report Improvements

robertisele:
- Persisted execution reports (CMEM-2618)
  - Added a configurable report manager that persists execution reports and allows to retrieve previous reports.
  - File and in-memory report managers are available.
  - There is a configurable retention time. Reports older than that will be deleted, if a new report is added.

- Error output for transformations (CMEM-2618)
  - Transformations have a new parameter "error output". When executing the workflow, all errornous entities together with the error description are written to the configured dataset.

- The following issues in the execution reports have been fixed (CMEM-2618):
  - Empty dataset reports
    - Dataset reports have been added for reading and writing datasets.   
  - Missing entities
    - Previously, if a URI did not generate a URI, the transformation silently did not generate any entity.
    - Now, those issues will be part of the transformation report validation issues.
  - Overwritten reports
    - If the same task has been added to a workflow multiple times, the workflow report only contained one task report. Multiple executions in the same workflow overwrote each other, so that only the last one was retained.
    - This is fixed and the report contains all operator executions.
  - Missing reports
    - The report only contained task reports for task types that generate a summary.
    - Now, the report contains all task reports. The reports are empty in the UI, we can think about displaying more information there.


andreas-schultz (approved):
Discovered issues will be moved to the follow up ticket.

PR #447, opened 2021-04-28
Improve workflow saving

robertisele:
- On loading, the save button is only enabled after the workflow has been loaded to prevent the user from saving an empty workflow.
  - A spinner is shown while the workflow is saved.
  - The save button is disabled while the workflow is saved.
  - A confirm dialog is shown, if the user tries to leave while a save is in progress.


chriswartner:
Works as described.

PR #449, opened 2021-05-06
Use latest OpenJDK 8 jre for Docker image

andreas-schultz:
- Use latest OpenJDK jre for Docker image
- Improve README regarding building the Docker image


jakubklimek:
Currently, when I do a fresh `docker pull`, I still get an image using java 8 update 151, which does not support TLS 1.3.

PR #518, opened 2021-12-02
Added endpoints to execute simple workflows asynchronously

robertisele:


silk-workbench/silk-workbench-workflow/app/controllers/workflowApi/WorkflowApi.scala
```diff
@@ -218,22 +220,191 @@ class WorkflowApi @Inject()() extends InjectedController with ControllerUtilsTra
...
+                                  )
+                                  instanceId: String): Action[AnyContent] = RequestUserContextAction { implicit request => implicit userContext =>
+    val (project, workflowTask) = getProjectAndTask[Workflow](projectName, workflowTaskName)
+
+    // Make sure that the activity has been successful
+    val activity = workflowTask.activity[WorkflowWithPayloadExecutor].instance(instanceId)
+    if(activity.status().isRunning) {
+      throw BadUserInputException("Workflow is still running.")
+    } else if(activity.status().failed) {
+      throw BadUserInputException("Workflow failed.", activity.status().exception)
```
andreas-schultz, line 399: Should this be rather a 5xx error instead of a 400 error? In the end it's not clear if the exception happened because of the user provided data or something internally.

silk-workbench/silk-workbench-workflow/app/controllers/workflowApi/WorkflowApi.scala
```diff
@@ -218,22 +220,191 @@ class WorkflowApi @Inject()() extends InjectedController with ControllerUtilsTra
...
+                                    in = ParameterIn.PATH,
+                                    schema = new Schema(implementation = classOf[String])
+                                  )
+                                  instanceId: String): Action[AnyContent] = RequestUserContextAction { implicit request => implicit userContext =>
+    val (project, workflowTask) = getProjectAndTask[Workflow](projectName, workflowTaskName)
+
+    // Make sure that the activity has been successful
+    val activity = workflowTask.activity[WorkflowWithPayloadExecutor].instance(instanceId)
+    if(activity.status().isRunning) {
+      throw BadUserInputException("Workflow is still running.")
```
andreas-schultz, line 397: I would prefer a 404 error here, since the result is not ready, yet.

PR #521, opened 2021-12-15
New workspace UI

andreas-schultz:


workspace/src/locales/manual/en.json
```diff
@@ -0,0 +1,612 @@
+{
```
haschek: Why do we have still 2 different language file for one language? This is still as difficult as before.

workspace/src/theme/grid.scss
```diff
@@ -0,0 +1,165 @@
+// utility
```
haschek: This `grid.scss` can probably be completely removed.

PR #527, opened 2022-01-11
New transform and linking editors

andreas-schultz:


workspace/src/app/store/ducks/common/typings.ts
```diff
@@ -60,7 +60,7 @@ export interface IOverviewArtefactItem {
 }
 
 /** The full task plugin description, including detailed schema. */
-export interface IDetailedArtefactItem {
+export interface IPluginDetails {
```
haschek, line 63: Typescript docs recommend not to use the I-prefix for interfaces. We should also stop it.

Some backgrounds:

* https://stackoverflow.com/questions/31876947/confused-about-the-interface-and-class-coding-guidelines-for-typescript/41967120#41967120
* https://developer.okta.com/blog/2019/06/25/iinterface-considered-harmful

PR #533, opened 2022-02-10
Feature/support key words and authorship chem 3717

arausly:


haschek (changes requested):

haschek:
One thing, names of `KeywordProp` and `KeywordProps` (defined as `KeywordProp[]`) are looking like they are not really correctly used, even if the code formally is valid.

Typescript usually recommends to use `*Props` as naming pattern for interfaces, so it shoud be `KeywordProps` instead of `KeywordProp`, and then just using `KeywordProps[]` for the tags array. Or it is `Keyword` and `Keywords` as type names.

`KeywordProp` and `KeywordProps` are just feeling not correctly used.

haschek (approved):

PR #543, opened 2022-03-31
Feature/activity overview cmem 2877

arausly:


workspace/src/app/views/pages/Project/ActivityInfoWidget.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { SERVE_PATH } from "../../../constants/path";
+
+const ActivityInfoWidget = () => {
+    const projectId = useSelector(commonSel.currentProjectIdSelector);
+    const dispatch = useDispatch();
+    const [t] = useTranslation();
+
+    const projectPath = `projects/${projectId}/activities?page=1&limit=25`;
+    return (
+        <Card>
```
haschek, line 24: Is there any reason why not to use the `Card` API to have a normal card header with options menu for the icon button here? At least pls create a follow up issue in JIRA for that widget.

haschek (changes requested):

haschek (approved):

PR #575, opened 2022-06-28
Move silk-react-components code to workspace code base and integrate components without i-frame

andreas-schultz:


andreas-schultz:
Corresponding gui-elements PR: https://github.com/eccenca/gui-elements/pull/63

PR #607, opened 2022-10-06
Fix/inline linking evalution not showing errors cmem 3331

arausly:


workspace/src/app/views/taskViews/transform/evalution/TransformRuleNodeEvaluation.tsx
```diff
@@ -0,0 +1,88 @@
+import { WhiteSpaceContainer, Tag, Tooltip } from "@eccenca/gui-elements";
```
andreas-schultz, line 1: Validation errors are currently not shown for transform rules. Can you change this?
Can you also delete `TransformRuleNodeEvaluation` and instead adapt `LinkRuleNodeEvaluation.tsx` to be used for transform and linking rules?

PR #619, opened 2022-10-22
Feature/support adding tags on task creation chem 4182

arausly:


workspace/src/app/views/shared/Metadata/MetadataUtils.tsx
```diff
@@ -48,6 +49,23 @@ export const queryTags = (projectId: string, filter?: string): Promise<FetchResp
...
+    taskId?: string
+): Promise<FetchResponse<IMetadata>> =>
+    projectId && taskId
+        ? fetch({
+              url: legacyApiEndpoint(`/projects/${projectId}/tasks/${taskId}/metadata`),
+              method: "put",
+              body: payload,
+          })
+        : fetch({
+              url: workspaceApi(`/projects/${projectId}/metaData`),
```
andreas-schultz, line 64: When both `projectId` and `taskId` are undefined, this will go against `/projects/undefined/...`.

andreas-schultz (approved):

PR #635, opened 2022-11-29
Variable workflow API improvements

andreas-schultz:
- support large input and output files
- support all file based datasets


silk-core/src/main/scala/org/silkframework/runtime/plugin/ClassPluginDescription.scala
```diff
@@ -142,16 +142,21 @@ object ClassPluginDescription {
...
+        categories = Seq(PluginCategories.uncategorized),
+        description = "",
+        documentation = "",
+        parameters = getParameters(pluginClass),
+        constructor = getConstructor(pluginClass),
+        pluginTypes = getPluginTypes(pluginClass)
+      )
+    } catch {
+      case ex: InvalidPluginException =>
+        throw new InvalidPluginException(s"Cannot extract plugin description for plugin class '${pluginClass.getCanonicalName}'. Details: ${ex.getMessage}")
```
robertisele, line 158: This will swallow nested causes and the original stacktrace. ex should be fed into the constructor as well.

PR #647, opened 2023-01-05
added hook to make search request on type

arausly:


workspace/src/app/views/shared/SearchBar/SearchBar.tsx
```diff
@@ -29,45 +30,24 @@ export function SearchBar({
...
-
-    const onClearanceHandler = () => {
-        setSearchInput("");
-        onSearch("");
-    };
-
-    const handleSearchEnter = () => {
-        onSearch(searchInput);
-    };
+    const { query, onChange, onEnter, onClear } = useSearch(onSearch);
```
andreas-schultz, line 39: The delay should be configurable, <= 0 for disable.

workspace/src/app/hooks/useSearch.tsx
```diff
@@ -0,0 +1,63 @@
...
+    onChange: (e: React.ChangeEvent<HTMLInputElement>) => void;
+    onClear: () => void;
+    onEnter: () => void;
+}
+
+export const useSearch = (onSearch: (query: string) => void, delay = 500): SearchHandlerReturnType => {
+    const [query, setQuery] = React.useState<string>("");
+    const [isSearching, setIsSearching] = React.useState<boolean>(false);
+
+    const debouncedSearchFunction = debounce(onSearch, delay);
```
andreas-schultz, line 17: Debounce not working as expected, i.e. for each character a query is fired.

workspace/src/app/views/shared/SearchBar/SearchBar.tsx
arausly: i'm guessing you mean >= 0, in that case, it's already added to the hook, where the default is 500 ms

PR #717, opened 2023-10-03
Feature/hotkey access to actions in di chem 4785

arausly:


workspace/src/app/views/shared/SearchBar/SearchBar.tsx
```diff
@@ -37,10 +44,13 @@ export function SearchBar({
     focusOnCreation = false,
     warnOfInvisibleCharacters = true,
     onEnter,
+    selectFirstResultItemOnEnter,
     ...otherProps
 }: IProps) {
     const [t] = useTranslation();
     const { query, setQuery, onChange, onEnter: onEnterRefreshSearch, onClear } = useSearch(onSearch, textQuery);
+    const workspaceSearchResult = useSelector(workspaceSel.resultsSelector);
```
andreas-schultz, line 52: Since the `SearchBar` is not only used in the workfpsace search, it would make sense to put the "select first item" logic into the onEnter callback instead.

PR #747, opened 2023-12-20
Support custom tasks with fixed output schema as input of transform and linking tasks

andreas-schultz:
- Support auto-completion of paths from the fixed schema in mapping and linking editors
- Support update of path caches when the custom task config changes


haschek (approved):
Looks good.

PR #749, opened 2023-12-21
Multiline editing of template values

arausly:


workspace/src/app/views/shared/modals/CreateArtefactModal/ArtefactForms/ArtefactFormParameter.tsx
```diff
@@ -194,7 +201,7 @@ export const ArtefactFormParameter = ({
                         maxWidth: showSwitchButton ? "calc(100% - 3.5px - 32px)" : "100%", // set full width minus tiny spacing and icon button width
                     }}
                 >
-                    {supportVariableTemplateElement && showVariableTemplateInput ? (
+                    {supportVariableTemplateElement && (showVariableTemplateInput || multiline) ? (
```
andreas-schultz, line 204: This would always enable templating mode even if no templating was selected by the user.

PR #781, opened 2024-05-28
Feature to add button to reload all caches CMEM-3850

arausly:


workspace/src/app/views/shared/TaskActivityOverview/TaskActivityOverview.tsx
```diff
@@ -105,6 +113,24 @@ export function TaskActivityOverview({ projectId, taskId }: IProps) {
...
+        .map((v) => `${v.lastUpdateTime}-${v.concreteStatus}-${v.progress}`)
+        .join("|");
+
+    React.useEffect(() => {
+        const currentlyRunningActivity = Array.from(activityStatusMap.values()).find((a) => a.isRunning);
+        const activity = cacheActivities[cursor.current];
+        //if no currently running activity then call the next cursor
+        if (!currentlyRunningActivity && activity) {
+            const activityFunctions = activityFunctionsCreator(activity);
+            activityFunctions.executeActivityAction("restart");
```
andreas-schultz, line 126: Currently this will display an error if the activity is already running. It should either restart the activity again after it has finished or ignore the error if it is OK that the previous restart suffices.

workspace/src/app/views/shared/TaskActivityOverview/TaskActivityOverview.tsx
```diff
@@ -105,6 +113,24 @@ export function TaskActivityOverview({ projectId, taskId }: IProps) {
         );
     };
 
+    const statusMapDependency = Array.from(activityStatusMap.values())
+        .map((v) => `${v.lastUpdateTime}-${v.concreteStatus}-${v.progress}`)
+        .join("|");
+
+    React.useEffect(() => {
+        const currentlyRunningActivity = Array.from(activityStatusMap.values()).find((a) => a.isRunning);
+        const activity = cacheActivities[cursor.current];
```
andreas-schultz, line 122: It would be better to start all activities at once, since there can be multiple long-running activities whose execution times will add up.

workspace/src/app/views/shared/TaskActivityOverview/TaskActivityOverview.tsx
```diff
@@ -105,6 +113,24 @@ export function TaskActivityOverview({ projectId, taskId }: IProps) {
         );
     };
 
+    const statusMapDependency = Array.from(activityStatusMap.values())
+        .map((v) => `${v.lastUpdateTime}-${v.concreteStatus}-${v.progress}`)
+        .join("|");
+
+    React.useEffect(() => {
+        const currentlyRunningActivity = Array.from(activityStatusMap.values()).find((a) => a.isRunning);
```
andreas-schultz, line 121: If there is already ANY running activity, e.g. an hour long execution of a transform task, the cache update won't start at all. This can probably be solved the point below to start all cache activities at once.

PR #785, opened 2024-06-06
Forward Workflow ID in context.

robertisele:


silk-core/src/main/scala/org/silkframework/runtime/plugin/PluginContext.scala
```diff
@@ -23,39 +23,44 @@ trait PluginContext {
 
   def templateVariables: TemplateVariablesReader
 
+  def workflowId: Option[Identifier]
```
andreas-schultz, line 26: Please add a comment, it is not clear what role a workflow plays in context of plugin creation and update.

robertisele: I have added comments to the attributes. Also made it clear that PluginContext is also used for plugin execution.

andreas-schultz (approved):

PR #807, opened 2024-07-30
Expose metrics for Prometheus using Micrometer

edufuga:
# Expose metrics for Prometheus using Micrometer

This PR adds support for exposing metrics using Micrometer, with the specific use-case of Prometheus in mind.

The main changes in this PR are:
1. Integrating the **functionality of Micrometer-based metrics** and configuring the Micrometer-based metrics to work with Prometheus as the monitoring and alerting software system
2. Exposing the **thread pool metrics** of the `Activity`
3. Exposing the **total number** of **projects** and **tasks** in a **workspace**
4. Exposing the number of **tasks** by their **specification** (Transform, Dataset, Linking, Task, Workflow).
5. Exposing the **runtime** of a **workflow**, **operator** and **dataset**, as well as the **number of entities** of a dataset.
6. Exposing the **version** (Workbench Core).
7. Add the **app name** as a metrics tag, via the property `metrics.tags`.
8. Add **descriptions** to the metrics.
9. Enable enabling/disabling the metrics functionality, via the property `metrics.enabled`.

### DIFF
--- a/silk-workspace/src/main/scala/org/silkframework/workspace/Project.scala
+++ b/silk-workspace/src/main/scala/org/silkframework/workspace/Project.scala
@@ -133,7 +133,7 @@ class Project(initialConfig: ProjectConfig, provider: WorkspaceProvider, val res
     * @throws NotFoundException
     */
   def activity(activityName: String): ProjectActivity[_ <: HasValue] = {
-    projectActivities.find(_.name == activityName)
+    projectActivities.find(_.name.toString == activityName)
       .getOrElse(throw NotFoundException(s"Project '$id' does not contain an activity named '$activityName'. " +
         s"Available activities: ${activities.map(_.name).mkString(", ")}"))
   }
--- a/silk-workspace/src/main/scala/org/silkframework/workspace/activity/workflow/LocalWorkflowExecutor.scala
+++ b/silk-workspace/src/main/scala/org/silkframework/workspace/activity/workflow/LocalWorkflowExecutor.scala
@@ -11,12 +11,13 @@ import org.silkframework.plugins.dataset.InternalDataset
 import org.silkframework.rule.TransformSpec
 import org.silkframework.runtime.activity.{ActivityContext, UserContext}
 import org.silkframework.runtime.metrics.MeterRegistryProvider
+import org.silkframework.runtime.metrics.MetricsConfig.prefix
 import org.silkframework.workspace.ProjectTask
 import org.silkframework.workspace.activity.transform.TransformTaskUtils._
 import org.silkframework.workspace.activity.workflow.ReconfigureTasks.ReconfigurableTask
 
 import java.util.logging.{Level, Logger}
-import scala.util.Try
+import scala.util.{Success, Try}
 import scala.util.control.NonFatal
 
 /**
@@ -54,7 +55,7 @@ case class LocalWorkflowExecutor(workflowTask: ProjectTask[Workflow],
       val registry = MeterRegistryProvider.meterRegistry
       val stopwatch: Timer.Sample = Timer.start()
       runWorkflow(context, updateUserContext(userContext))
-      stopwatch.stop(registry.timer("workflow.execution", "workflow", workflowTask.id.toString))
+      stopwatch.stop(registry.timer(s"$prefix.workflow.execution", "workflow", workflowTask.id.toString))
     } catch {
       case cancelledWorkflowException: StopWorkflowExecutionException if !cancelledWorkflowException.failWorkflow =>
         // In case of an cancelled workflow from an operator, the workflow should still be successful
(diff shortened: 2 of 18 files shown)


silk-workspace/src/main/scala/org/silkframework/workspace/Project.scala
```diff
@@ -133,7 +133,7 @@ class Project(initialConfig: ProjectConfig, provider: WorkspaceProvider, val res
     * @throws NotFoundException
     */
   def activity(activityName: String): ProjectActivity[_ <: HasValue] = {
-    projectActivities.find(_.name == activityName)
+    projectActivities.find(_.name.toString == activityName)
```
edufuga, line 136: This has nothing to do with the metrics themselves, but I think it was a bug.

silk-workspace/src/main/scala/org/silkframework/workspace/activity/workflow/LocalWorkflowExecutor.scala
```diff
@@ -179,6 +187,8 @@ case class LocalWorkflowExecutor(workflowTask: ProjectTask[Workflow],
             writeErrorOutput(operatorNode, executorOutput)
             process(result)
           }
+          stopwatch.stop(registry.timer(s"$prefix.workflow.operator.execution", "operator", operator.nodeId))
```
robertisele, line 190: Unfortunately, the node id is not unique between workflows so it's not clear, which workflow the resulting metric value belongs to. Is it possible to add two tags here? One for the workflow and another one for the operator?

silk-workspace/src/main/scala/org/silkframework/workspace/activity/workflow/LocalWorkflowExecutor.scala
```diff
@@ -293,9 +303,16 @@ case class LocalWorkflowExecutor(workflowTask: ProjectTask[Workflow],
                                        (implicit workflowRunContext: WorkflowRunContext): Unit = {
     val resolvedDataset = resolveDataset(datasetTask(workflowDataset), replaceSinks)
     try {
+      val registry = MeterRegistryProvider.meterRegistry
+      val stopwatch: Timer.Sample = Timer.start()
       executeAndClose("Writing", workflowDataset.nodeId, resolvedDataset, Seq(entityTable), ExecutorOutput.empty) { _ =>
         // ignore result
       }
+      stopwatch.stop(registry.timer(s"$prefix.workflow.dataset.execution", "dataset", workflowTask.id.toString))
```
robertisele, line 311: The operator metric uses the node Id, so I think it should be used here as well (workflowDataset.nodeId) instead of the task id.

silk-workspace/src/main/scala/org/silkframework/workspace/activity/workflow/LocalWorkflowExecutor.scala
```diff
@@ -168,7 +174,9 @@ case class LocalWorkflowExecutor(workflowTask: ProjectTask[Workflow],
 
       if (!cancelled && !isExecutedWorkflow) {
         try {
-          executeAndClose("Executing", operatorNode.nodeId, operatorTask, inputResults.flatten, executorOutput) { result =>
+          val registry = MeterRegistryProvider.meterRegistry
+          val stopwatch: Timer.Sample = Timer.start()
+          val result = executeAndClose("Executing", operatorNode.nodeId, operatorTask, inputResults.flatten, executorOutput) { result =>
```
robertisele, line 179: This would not include the execution time of failed workflows. That's probably fine, but we should ask for feedback if this is expected.

silk-workspace/src/main/scala/org/silkframework/workspace/activity/workflow/LocalWorkflowExecutor.scala
```diff
@@ -49,7 +52,10 @@ case class LocalWorkflowExecutor(workflowTask: ProjectTask[Workflow],
                   (implicit userContext: UserContext): Unit = {
     cancelled = false
     try {
+      val registry = MeterRegistryProvider.meterRegistry
+      val stopwatch: Timer.Sample = Timer.start()
       runWorkflow(context, updateUserContext(userContext))
+      stopwatch.stop(registry.timer(s"$prefix.workflow.execution", "workflow", workflowTask.id.toString))
```
robertisele, line 58: This would not include the execution time of failed workflows. That's probably fine, but we should ask for feedback if this is expected.

silk-workspace/src/main/scala/org/silkframework/workspace/activity/workflow/LocalWorkflowExecutor.scala
edufuga: I added the **workflow (execution) ID** as a new tag pair.

silk-workspace/src/main/scala/org/silkframework/workspace/activity/workflow/LocalWorkflowExecutor.scala
edufuga: I added the **workflow dataset node ID** as a new tag pair.

robertisele: Now I get: `cmem_workflow_dataset_execution_seconds_max{app="di",dataset="Workflow_a01db34855e3bb9e",} 0.0642097`. Shouldn't this also include both the workflow and the node id?

edufuga: My bad. I changed the one below it. It should be fine now.

PR #818, opened 2024-09-05
Changed codemirror libraries and made corresponding refactors

arausly:


workspace/src/app/views/shared/RuleEditor/view/ruleNode/RuleParameterInput.tsx
```diff
@@ -20,7 +20,7 @@ import { TextAreaWithCharacterWarnings } from "../../../extendedGuiElements/Text
 import { IPropertyAutocomplete } from "@ducks/common/typings";
 import { LanguageFilterProps, PathInputOperator } from "./PathInputOperator";
 import { supportedCodeRuleParameterTypes } from "../../RuleEditor.typings";
-import { SupportedCodeEditorModes } from "@eccenca/gui-elements/src/extensions/codemirror/CodeMirror";
+import { SupportedCodeEditorModes } from "@eccenca/gui-elements/src/extensions/codemirror/hooks/useCodemirrorModeExtension.hooks";
```
haschek, line 23: Always if we need an import that is not available via `@eccenca/gui-elements` base without adding sub paths, then we should think if we do it correct in gui elements.

In this case you could also change it to importing `CodeEditorProps` and use `CodeEditorProps["mode"]`.

workspace/src/app/views/taskViews/linking/activeLearning/config/ManualComparisonPairSelection.tsx
```diff
@@ -313,6 +313,7 @@ const PathAutoCompletion = ({
         <ComparisonDataCell className="diapp-linking-learningdata__pathselection">
             {resetting ? null : (
                 <AutoSuggestion
```
haschek, line 315: `AutoSuggestion` is deprecated, pls use the new component name `CodeAutocompleteField`.

haschek (changes requested):

workspace/package.json
```diff
@@ -35,8 +35,7 @@
         "@uppy/xhr-upload": "1.7.5",
         "axios": "^0.25.0",
         "bfj": "6.1.1",
-        "codemirror": "^5.52.2",
-        "codemirror-formatting": "1.0.0",
+        "codemirror": "^6.0.1",
```
haschek, line 38: Why do we need `codemirror` lib as depency in Silk? Why is `CodeEditor` from the gui elements not enough?

haschek (changes requested):

PR #831, opened 2024-10-11
Add optional width to node layout.

robertisele:


workspace/src/app/views/taskViews/transform/TransformRuleEditor.tsx
```diff
@@ -158,8 +158,13 @@ export const TransformRuleEditor = ({
...
                 node.position = {
-                    x: pos[0],
-                    y: pos[1],
+                    x: pos.x,
+                    y: pos.y,
+                };
+                node.dimension = {
+                    ...node.dimension,
+                    width: pos.width,
+                    height: pos.height,
```
andreas-schultz, line 167: Why is the height saved and restored if the user cannot change it? Can this lead to problems in combination with advanced parameters?

andreas-schultz (changes requested):

workspace/src/app/views/shared/RuleEditor/view/ruleNode/RuleNodeMenu.tsx
```diff
@@ -103,6 +111,13 @@ export const RuleNodeMenu = ({
                         )}
                     />
                 ) : null}
+                <MenuItem
```
haschek, line 114: Please update the node menu accordion our latest spec, moving the "remove" menu item to the bottom and separate it by a divider.

PR #954, opened 2025-08-26
Use bezier edges in workflow editor (CMEM-6721)

haschek:


haschek:
We won't use bezier edges as default.

PR #960, opened 2025-09-10
Feature/sql update cmem 4031

chriswartner:


silk-core/src/main/scala/org/silkframework/dataset/sql/SqlDataset.scala
```diff
@@ -0,0 +1,8 @@
+package org.silkframework.dataset.sql
+
+import org.silkframework.dataset.Dataset
+
+// TODO implemented by Snowflake, JDBC
```
robertisele, line 5: Looks like this TODO is done and can be removed?

PR #969, opened 2025-10-02
Forward to finalized next major version of GUI elements (CMEM-6943)

haschek:


workspace/src/theme/index.scss
```diff
@@ -10,7 +10,7 @@ $eccgui-selector-text-spot-highlight: "[id^=parameter_doc_]";
  */
 
 // Transformation editor
-@import "./../app/views/pages/MappingEditor/style/style.scss";
+//@import "./../app/views/pages/MappingEditor/style/style.scss";
```
andreas-schultz: Can this be removed?

PR #1021, opened 2026-03-23
Add Jinja templating support to SPARQL operators

robertisele:


silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/tasks/templating/SparqlJinjaTemplate.scala
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  override lazy val outputSchema: EntitySchema = {
+    val vars = SparqlSelectVarExtractor.extractSelectVars(rawTemplate)
+    val paths = vars.map(v => TypedPath(UntypedPath(v), ValueType.STRING, isAttribute = false))
+    EntitySchema(typeUri = Uri(""), typedPaths = paths.toIndexedSeq)
+  }
+
+  override def isStaticTemplate: Boolean = {
+    entityPropertyNames.isEmpty
```
andreas-schultz: This will be false when only using e.g. {{ input.config.graph }}.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/tasks/SparqlSelectCustomTask.scala
```diff
@@ -50,38 +51,82 @@ case class SparqlSelectCustomTask(
...
+      }
+    } else {
+      FixedNumberOfInputs(Seq(FixedSchemaPort(SparqlEndpointEntitySchema.schema)))
+    }
   }
 
   override def outputPort: Option[Port] = {
-    Some(FixedSchemaPort(outputSchema))
+    if (outputSchema.typedPaths.isEmpty) {
+      Some(UnknownSchemaPort)
```
andreas-schultz, line 112: This is currently problematic. If the result variables cannot be inferred from the template, an unknown schema is defined. However, the SPARQL Select executor is based on the output schema of the task. It should probably rely on the entity schema of the input entities instead.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/tasks/SparqlUpdateCustomTask.scala
```diff
@@ -32,40 +34,30 @@ case class SparqlUpdateCustomTask(
...
+  templatingMode: String = "jinja"
 ) extends CustomTask {
   assert(batchSize >= 1, "Batch size must be greater zero!")
 
-  val templatingEngine: SparqlUpdateTemplatingEngine = templatingMode match {
-    case SparqlUpdateTemplatingMode.simple => SparqlUpdateTemplatingEngineSimple(sparqlUpdateTemplate.str, batchSize)
-    case SparqlUpdateTemplatingMode.velocity => SparqlTemplatingEngineVelocity(sparqlUpdateTemplate.str, batchSize)
+  val compiledTemplate: SparqlTemplate = SparqlTemplate.create(templatingMode, sparqlUpdateTemplate.str)
+  for(variables <- sparqlUpdateTemplate.variables) {
+    compiledTemplate.validate(variables, None)
```
andreas-schultz: The batch size is always set to None instead of the actual batch size.

silk-plugins/silk-plugins-rdf/src/test/scala/org/silkframework/plugins/dataset/rdf/executors/LocalSparqlSelectExecutorTest.scala
```diff
@@ -58,14 +63,101 @@ class LocalSparqlSelectExecutorTest extends AnyFlatSpec
...
+    val entities = LocalSparqlSelectExecutor().executeOnSparqlEndpoint(task, taskWithEndpoint(sparqlEndpoint), None, limit, Some(reportUpdater))
     entities.headOption // Needed to actually execute the query
     correctTimeout mustBe true
   }
 
-  private def sparqlEndpointStub(selectCallback: SparqlEndpoint => Unit = _ => {}): SparqlEndpoint = {
+  it should "generate one query per input entity when useDefaultDataset is set and the template references entity values" in {
+    val query = """SELECT ?p ?o WHERE { <{{ input.entity.s }}> ?p ?o }"""
+    val rowsPerQuery = 2
+    val task = SparqlSelectCustomTask(query, limit = rowsPerQuery.toString, useDefaultDataset = true)
```
andreas-schultz, line 79: This should probably be documented in the limit parameter. Currently it says:

> If set to a positive integer, the number of results is limited

It should probably be say something like "per query"

silk-plugins/silk-plugins-rdf/src/main/resources/org/silkframework/plugins/dataset/rdf/tasks/SparqlSelectCustomTask.md
```diff
@@ -1,22 +1,125 @@
...
-The SPARQL Select query plugin is an example of a _RDF task_ or _operator_. Such a task can be used in a workflow,
-connecting an input to an output. In this specific case, the _input_ is — in essence — a _SPARQL endpoint_ and the
-_output_ is the entity table containing the _SPARQL results_ of the SPARQL SELECT query execution.
+The _input_ depends on the configuration:
 
-In general terms, a [SPARQL 1.1 SELECT](https://www.w3.org/TR/sparql11-query/#select) query is supported. One of the
-simplest examples is `SELECT * WHERE { ?s ?p ?o }`.
+- By default, the query is executed against the connected input, which must be a _SPARQL endpoint_
+  (i.e. an RDF dataset).
+- When **Use default RDF dataset** (`useDefaultDataset`) is enabled, the query is executed against the project's
```
andreas-schultz: Why does it say the "project's default RDF dataset"? The only seems to be a global default dataset.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/tasks/templating/SparqlJinjaTemplate.scala
robertisele: I am not sure what the original purpose of this method was. Since it's used to derive the input schema, I changed it's name to `requiresInput `and updated the logic so it's only true, if `input `is accessed. Also updated the test.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/tasks/SparqlUpdateCustomTask.scala
robertisele: Fixed

silk-plugins/silk-plugins-rdf/src/test/scala/org/silkframework/plugins/dataset/rdf/executors/LocalSparqlSelectExecutorTest.scala
robertisele: I updated the description of the limit parameter.

silk-plugins/silk-plugins-rdf/src/main/resources/org/silkframework/plugins/dataset/rdf/tasks/SparqlSelectCustomTask.md
robertisele: I updated the doc, which was still from an earlier version.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/tasks/SparqlSelectCustomTask.scala
robertisele: In case, the output schema is empty (i.e. parsing failed), the best I could come up with to fix this is to infer the actual schema from the variables of the first SPARQL result. But we can discuss this if you have another idea.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/tasks/SparqlSelectCustomTask.scala
```diff
@@ -40,48 +35,99 @@ import scala.util.Try
...
+  }
+
+  def requiresInput: Boolean = queryTemplate.requiresInput
+
+  def expectedInputSchema: EntitySchema = queryTemplate.inputSchema
+
   override def inputPorts: InputPorts = {
-    FixedNumberOfInputs(Seq(FixedSchemaPort(SparqlEndpointEntitySchema.schema)))
+    if (useDefaultDataset) {
+      if (requiresInput) {
```
andreas-schultz, line 100: There might be an edge case here when `requiresInput` is set when `input.config` is used,
useDefaultDataset == true and there is no `input.entity.*` usage.
This seems to have the effect in the `LocalSparqlSelectExecutor` that no query will be executed.

robertisele: I added two tests to LocalSparqlSelectExecutorTest and fixed this behaviour.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/tasks/SparqlSelectCustomTask.scala
andreas-schultz: This seems to be the only possibility given that there is no per-entity schema.

PR #1033, opened 2026-04-09
Add In-workflow dataset

robertisele:


silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/endpoint/JenaModelEndpoint.scala
```diff
@@ -1,17 +1,33 @@
...
 
 /**
  * A SPARQL endpoint which executes all queries on a Jena Model.
  */
 class JenaModelEndpoint(model: Model) extends JenaEndpoint {
 
+  private val byteLimit: Long = Resource.maxInMemorySize()
+  @volatile private var estimatedBytesWritten: Long = 0L
+
+  model.register(new StatementListener {
```
andreas-schultz: This only keeps track of memory footprint of the data written with this instance of the endpoint. When consecutive endpoints write to the same Model, the accumulated data can exceed the max size.

silk-plugins/silk-plugins-rdf/src/main/resources/org/silkframework/plugins/dataset/rdf/datasets/InMemoryDataset.md
```diff
@@ -44,33 +48,58 @@ All three sinks ultimately write into the same in-memory graph; there is no sepa
...
-- **Parameter:** `Clear graph before workflow execution` (boolean)
+- **Parameter:** `workflowScoped` (boolean)
 - **Default:** `true`
 
-Behaviour:
+When `true` (default, workflow-scoped mode):
+- Data is stored in a separate in-memory graph for each workflow execution.
+- Concurrent workflow executions are fully isolated from each other.
+- A dataset task in a nested workflow shares the same graph as the parent for the same task identifier. Data written by the parent is available in the nested workflow and vice versa.
+- If the dataset is read from outside a workflow context, the data from the most recently started executor is returned.
```
andreas-schultz: This does not seem correct. In the current implementation this seems to be more accurate:

If the dataset is read from outside a workflow context, the data of the workflow execution that most recently accessed this dataset is returned.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/datasets/InMemoryDatasetExecutor.scala
```diff
@@ -0,0 +1,54 @@
...
+  @volatile private var initialized: Boolean = false
+  @volatile private var modelKey: Option[ExecutionModelKey] = None
+  @volatile private var plugin: Option[InMemoryDataset] = None
+
+  override def access(task: Task[DatasetSpec[InMemoryDataset]], execution: LocalExecution): DatasetAccess = {
+    val datasetPlugin = task.data.plugin
+    if (datasetPlugin.workflowScoped) {
+      if (!initialized) {
+        initialized = true
+        model = execution.parentExecution
```
andreas-schultz: This does not seem to cover the following case:

- Nested workflow accesses the dataset first and writes to it.
- Parent workflow reads from the dataset.

The parent workflow should not see the written data.

silk-plugins/silk-plugins-rdf/src/test/scala/org/silkframework/plugins/dataset/rdf/datasets/InMemoryDatasetWorkflowScopedTest.scala
```diff
@@ -0,0 +1,171 @@
...
+    val parentEndpoint = sparqlEndpoint(parentExecutor.access(parentTask, parentExecution))
+    parentEndpoint.update("INSERT DATA { <http://s> <http://p> <http://o> }")
+
+    val childExecution = LocalExecution(false, workflowId = Some(Identifier("childWf")), parentExecution = Some(parentExecution))
+    val childExecutor = new InMemoryDatasetExecutor()
+    val childEndpoint = sparqlEndpoint(childExecutor.access(childTask, childExecution))
+    childEndpoint.select(tripleCountQuery).bindings.size mustBe 0
+  }
+
+  it should "clean up model on close()" in {
```
andreas-schultz: Clean up mode implies that the model is cleared. But it is only removed from the lookup.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/datasets/InMemoryDataset.scala
```diff
@@ -25,33 +31,75 @@ import org.silkframework.runtime.plugin.annotations.{Param, Plugin, PluginRefere
...
+    executionModels.remove(key)
+
+  /** Switches [[model]] to the given execution's model so out-of-workflow reads see current data. */
+  private[datasets] def updateData(newModel: Model): Unit =
+    model = newModel
+
+  // In workflow-scoped mode the executor owns the model lifecycle, so sinks must not drop the graph.
+  private def dropGraph: Boolean = !workflowScoped && clearGraphBeforeExecution
+
+  override def sparqlEndpoint: SparqlEndpoint = new JenaModelEndpoint(model)
```
andreas-schultz: Not 100% sure, but shouldn't tasks like the `SPARQL Select query` task access this? I.e. they will always get the "external" model instead of the workflow execution model?

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/endpoint/JenaModelEndpoint.scala
```diff
@@ -1,17 +1,33 @@
...
+    override def addedStatement(s: Statement): Unit =
```
andreas-schultz: The listener does not seem to be removed from the model when the endpoint is discarded. This accumulates listeners with every JenaModelEndpoint. The endpoint itself is also kept and not garbage collected. I've added a JMH benchmark `InMemoryDatasetEndpointBenchmark.scala` which can be executed like this:

```
sbt 'project pluginsRdf' 'Jmh/run org.silkframework.plugins.dataset.rdf.InMemoryDatasetEndpointBenchmark'
```

This shows the effect. Changing data in the model takes more and more time the more endpoints were created.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/endpoint/JenaModelEndpoint.scala
robertisele: Good catch! Not to make this more complicated, I introduced a new InMemoryJenaModelEndpoint that's now directly referenced by the InMemoryDataset and enforces the limit. In the future, we can use this elsewhere or make it more flexible.

silk-plugins/silk-plugins-rdf/src/main/resources/org/silkframework/plugins/dataset/rdf/datasets/InMemoryDataset.md
robertisele: I updated the doc.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/datasets/InMemoryDatasetExecutor.scala
robertisele: Also good catch! I improve the logic to cover this case and added a test.

silk-plugins/silk-plugins-rdf/src/test/scala/org/silkframework/plugins/dataset/rdf/datasets/InMemoryDatasetWorkflowScopedTest.scala
robertisele: I updated the test and named it better. It does also remove the pointer to the model now on close, but not the one in the InMemoryDataset itself, since we want this to stay accessible.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/datasets/InMemoryDataset.scala
robertisele: I am not sure what you mean. This is probably obsolete now with the new InMemoryJenaModelEndpoint?

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/endpoint/JenaModelEndpoint.scala
robertisele: This is fixed now since only a single InMemoryJenaModelEndpoint with a single model is used without re-registering new listeners on every access.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/datasets/InMemoryDataset.scala
andreas-schultz: The `LocalSparqlSelectExecutor` does not use the `access` path of the dataset, but the `sparqlEndpoint` directly. E.g. when multiple instances of a workflow using this dataset run in parallel, the sparql select executor could get a version of another workflow.  Same for LocalSparqlCopyExecutor and SPARQL update execution.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/datasets/InMemoryDataset.scala
```diff
@@ -25,33 +30,85 @@ import org.silkframework.runtime.plugin.annotations.{Param, Plugin, PluginRefere
...
 
   /**
-    * Returns a entity sink for writing entities to the data set.
-    */
-  override def entitySink(implicit userContext: UserContext): EntitySink = new SparqlSink(SparqlParams(), sparqlEndpoint, dropGraphOnClear = clearGraphBeforeExecution)
+   * Endpoints for all current workflow executions, keyed by [[ExecutionModelKey]].
+   * Uses a WeakHashMap so entries are automatically cleaned up by GC when the key is no longer referenced.
+   * Entries are also explicitly removed by [[InMemoryDatasetExecutor.close()]] when the execution finishes.
+   */
+  private val executionEndpoints: java.util.Map[ExecutionModelKey, InMemoryJenaModelEndpoint] =
```
andreas-schultz, line 63: This approach using a weak hashmap might lead to the following bug, especially when there is high memory pressure:

1. A nested workflow accsesses the dataset first, generates a key (equal keys are always re-created and are not the same object) for the hashmap and finishes.
2. The executors of the nested workflow are garbage collected -> the `modelKey` strong reference dissappears, since it is part of the in-memory dataset executor.
3. The entry in the weak hashmap is removed.
4. The main workflow accesses the in-memory dataset and creates a new endpoint because the entry is missing.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/datasets/InMemoryDataset.scala
robertisele: I changed this and updated `LocalSparqlSelectExecutpr `and others to use `RdfDatasetAccess.forExecution` to obtain  the SPARQL endpoint through the Executor.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/datasets/InMemoryDataset.scala
robertisele: I fixed this and added a new test for this to `InMemoryDatasetWorkflowScopedTest`

PR #1037, opened 2026-04-27
Split the execution of a transformer into a separate class.

robertisele:


silk-workbench/silk-workbench-rules/app/controllers/transform/EvaluateTransformApi.scala
```diff
@@ -230,11 +230,11 @@ class EvaluateTransformApi @Inject()(implicit accessMonitor: WorkbenchAccessMoni
 
     val ruleSchema = ruleSchemaById(task, parentRuleId)
     val inputSchema = ruleSchema.inputSchema.copy(typedPaths = transformRule.sourcePaths.toIndexedSeq)
-    val ruleWithContext = transformRule.withContext(task.taskContext)
 
     val entities = task.dataSource.retrieve(inputSchema, Some(limit)).entities.take(limit)
+    val ruleExec = transformRule.execution()
```
andreas-schultz: There is no task context supplied, evaluation will fail for operators that use the task context.

silk-rules/src/main/scala/org/silkframework/rule/TransformRule.scala
```diff
@@ -140,13 +126,39 @@ sealed trait TransformRule extends Operator with HasMetaData {
     }
   }
 
-  override def withContext(taskContext: TaskContext): TransformRule = this
+  override def execution(taskContext: TaskContext = TaskContext.empty): TransformRuleExecution
```
andreas-schultz, line 129: Why is there a default value. Shouldn't it be explicit about when a task context is not needed?

silk-workbench/silk-workbench-rules/app/controllers/linking/LinkingTaskApi.scala
```diff
@@ -1159,6 +1159,7 @@ class LinkingTaskApi @Inject() (accessMonitor: WorkbenchAccessMonitor) extends I
     if (evaluationActivity.control.status.get.isEmpty) {
       evaluationActivity.control.startBlocking()
     }
+    val ruleExec = linkTask.data.rule.execution()
```
andreas-schultz: Here, no task context is supplied.

silk-plugins/silk-serialization-json/src/main/scala/org/silkframework/serialization/json/LinkingSerializers.scala
```diff
@@ -64,7 +64,7 @@ object LinkingSerializers {
         case _ =>
           // Link does not provide evaluation details, need to evaluate it first
           for (linkingRule <- rule; entities <- link.entities) {
-            val details = DetailedEvaluator(linkingRule, entities).details
+            val details = DetailedEvaluator(linkingRule.execution(), entities).details
```
andreas-schultz: Here the task context is also missing.

silk-rules/src/main/scala/org/silkframework/rule/execution/EvaluateTransform.scala
```diff
@@ -24,12 +24,13 @@ class EvaluateTransform(source: DataSource,
   private var cachedValues = Seq[DetailedEntity]()
 
   def execute()(implicit context: PluginContext): Seq[DetailedEntity] = {
+    val ruleExecs = rules.map(_.execution())
```
andreas-schultz: Here, the task context should probably also be supplied.

silk-workbench/silk-workbench-rules/app/controllers/transform/EvaluateTransformApi.scala
robertisele: fixed

silk-rules/src/main/scala/org/silkframework/rule/TransformRule.scala
robertisele: You are right, I removed the default and set it manually at all places.

silk-workbench/silk-workbench-rules/app/controllers/linking/LinkingTaskApi.scala
robertisele: fixed

silk-plugins/silk-serialization-json/src/main/scala/org/silkframework/serialization/json/LinkingSerializers.scala
robertisele: fixed

silk-rules/src/main/scala/org/silkframework/rule/execution/EvaluateTransform.scala
robertisele: fixed

PR #1038, opened 2026-04-28
Support hidden search tags for tasks

andreas-schultz:
- exact match only
- case-insensitive
- not shown in the UI


workspace/src/app/views/shared/modals/RecentlyViewedModal.tsx
haschek, line 133: Not a change from the current feature addition but should this log message be here?

haschek, line 182: Should we add a `hiddenSearchTagsRenderer` here that dislays only those hidden tags that have a query match, plus the match highlighting.

haschek (changes requested):

workspace/src/app/views/shared/modals/RecentlyViewedModal.tsx
andreas-schultz: The idea was that they are not displayed, hence hidden. It should be clear though, why a result is returned. In the concrete use case the hidden string was just another version of a displayed tag plus the selectivity is very low, i.e. there is probably only one result returned for that string. When to use those hidden search tags is explained in the backend code.

andreas-schultz: It's probably a leftover from debugging and could be removed.

andreas-schultz: This will be removed in another issue.

haschek (approved):

PR #1040, opened 2026-04-30
Still support undo/redo after saving a mapping/linking rule in the editors

andreas-schultz:
- Allow to reset to last saved state. If possible without clearing the undo/redo queue.


J-x-D (approved):
lgtm :)

PR #1043, opened 2026-05-07
Reusable rule blocks

andreas-schultz:


silk-rules/src/main/scala/org/silkframework/rule/RuleBlockSpec.scala
```diff
@@ -0,0 +1,282 @@
...
+import scala.language.implicitConversions
+import scala.xml.{Elem, Node, Null, PCData}
+
+@Plugin(
+  id = "ruleBlock",
+  label = "Rule block",
+  categories = Array("Transform"),
+  description = "Defines a reusable transform rule block that can be referenced from transform and linking rules."
+)
+case class RuleBlockSpec(@Param(label = "Rule block model",
```
J-x-D, line 28: Since you introduce this new `RuleBlockSpec` extending `TaskSpec`, you probably have to add it to the ProjectMarshallingTrait.exportProject function:
```
// Export tasks
    exportTasks[DatasetSpec[Dataset]](project, outputWorkspaceProvider, resources, exportUserData)
    exportTasks[TransformSpec](project, outputWorkspaceProvider, resources, exportUserData)
    exportTasks[LinkSpec](project, outputWorkspaceProvider, resources, exportUserData)
    exportTasks[Workflow](project, outputWorkspaceProvider, resources, exportUserData)
    exportTasks[CustomTask](project, outputWorkspaceProvider, resources, exportUserData)
```

silk-workspace/src/main/scala/org/silkframework/workspace/Project.scala
```diff
@@ -66,6 +66,7 @@ class Project(initialConfig: ProjectConfig, provider: WorkspaceProvider, val res
     // Register all default modules
     registerModule[DatasetSpec[Dataset]]()
     registerModule[TransformSpec]()
+    registerModule[RuleBlockSpec]()
```
J-x-D, line 69: I think we should add a validator here as well. The frontend warns about the impact of dropping a RuleBlock, but a REST operation (which, in the future, could be performed using cmemc) would drop it, breaking the entire operator subtree. `WorkflowValidator` (two lines below) could work as a reference.

silk-rules/src/main/scala/org/silkframework/rule/RuleBlockSpec.scala
```diff
@@ -0,0 +1,282 @@
...
+    }
+  }
+}
+
+case class RuleBlockPort(id: Identifier = Operator.generateId,
+                         label: String = "",
+                         description: String = "",
+                         displayOrder: Int = 1,
+                         deprecated: Boolean = false) {
+  assert(label.nonEmpty, "Label of rule block port must not be empty!")
```
J-x-D: I think we should throw a `ValidationException` here. Otherwise, if I'm correct, we throw a 5xx HTTP error instead of a 4xx.

silk-rules/src/main/scala/org/silkframework/rule/RuleBlockSpec.scala
```diff
@@ -0,0 +1,282 @@
...
+      case (id, duplicates) if duplicates.size > 1 => id
+    }.toSeq
+    val duplicateDisplayOrders = ports.map(_.displayOrder).groupBy(identity).collect {
+      case (displayOrder, duplicates) if duplicates.size > 1 => displayOrder
+    }.toSeq.sorted
+    if(duplicatePortIds.nonEmpty) {
+      throw new ValidationException(s"Duplicate rule block port IDs found: ${duplicatePortIds.mkString(", ")}")
+    }
+    if(duplicateDisplayOrders.nonEmpty) {
+      throw new ValidationException(s"Duplicate rule block port display orders found: ${duplicateDisplayOrders.mkString(", ")}")
```
J-x-D, line 80: `JsonSerializers`and `TaskSerializers` set the `displayOrder` to 1. If users use multiple serializers without a self-set `displayOrder`, this exception is thrown. I assume users would be confused by the error message. Thus, we could either improve the error message or use an index-based default value in `JsonSerializers`and `TaskSerializers`.

silk-workspace/src/main/scala/org/silkframework/workspace/PluginUsage.scala
J-x-D, line 57: I think we should add another case for the RuleBlockSpec here as well. Otherwise, we don't list deprecated plugins used inside the RuleBlocks.

workspace/src/app/views/taskViews/linking/evaluation/tabView/LinkingEvaluationTabView.tsx
```diff
@@ -180,18 +182,25 @@ const LinkingEvaluationTabView: React.FC<LinkingEvaluationTabViewProps> = ({ pro
...
-                        linkingTaskId,
-                        pagination,
-                        searchQuery,
-                        filters,
-                        linkSortBy,
-                        showReferenceLinks,
-                        !showReferenceLinks,
-                    )
-                )?.data;
+                const [results, summaries] = await Promise.all([
```
J-x-D: Arrays in `Promise.all()` actually run sequentially. Thus, if `requestRuleBlockSummaries` fails, it kills the whole tab. It further fetches this data on every re-render. I think a separate `useEffect` that listens for `projectId` makes more sense here.

J-x-D:
Screenshot 2026-07-02 at 12 41 43

Using a RuleBlock works inside a Transform task. Once I use that Transform task inside a Linking task, I run into issues.

Steps to reproduce:

1. Create a project and upload a dataset.
2. Create a rule block (I just used the lower-case operator).
3. Create a transform task with the previously uploaded dataset as the source.
4. In a value mapping, use the rule block operator and evaluate the transformation (this works).
5. Create a Linking task and select the Transform task as source input; I just used the previously uploaded dataset as a target. 
6. I then used the equality operator (which you can see in the screenshot). 
7. Run the linking evaluation
-> The activity fails, as you can see in the screenshot. 

That could be related to caching issues.

J-x-D:
Screenshot 2026-07-02 at 13 54 32

Here, the edges are broken. 
I created the RuleBlock first, then used it in a transformation, and went back to edit it. That's what I see without editing anything.


FYI: If I open the preview from a transformation, the edges are visible.

J-x-D:
I first created a RuleBlock containing two ports and lowercase + uppercase operators. I then deleted a port and went back to my previously created transformation that uses the RuleBlock. It still shows the two inputs (you know, those circles with the edges) and doesn't show me an error. I don't see examples after evaluating the subtree (menu item click), which is as expected. 

If I save the corrupt transformation I further don't receive an error message.

workspace/src/app/views/shared/RuleEditor/view/ruleNode/RuleNodeMenu.tsx
J-x-D, line 111: Inside the RuleBlock canvas, it doesn't make sense to evaluate the subtree, since users can't specify actual input data.

silk-rules/src/main/scala/org/silkframework/rule/LinkSpec.scala
```diff
@@ -135,6 +136,8 @@ case class LinkSpec(@Param(label = "Source input", value = "The source input to
 
   override def outputTasks: Set[Identifier] = output.value.toSet
 
+  override def referencedTasks: Set[Identifier] = inputTasks ++ outputTasks ++ referencedRuleBlocks
```
robertisele, line 140: The Scaladoc in the parent trait TaskSpec needs to be updated, since it now contains not just input and output tasks.

silk-core/src/main/scala/org/silkframework/runtime/plugin/TaskResolver.scala
```diff
@@ -0,0 +1,47 @@
...
+    }
+  }
+
+  def fromProject(project: ProjectTrait)(implicit userContext: UserContext): TaskResolver = {
+    ProjectTraitTaskResolver(project, userContext)
+  }
+
+  private case class ProjectTraitTaskResolver(project: ProjectTrait, implicit private val userContext: UserContext) extends TaskResolver {
+
+    override def resolve(id: Identifier): Task[_ <: TaskSpec] = project.task(id)
```
robertisele: This probably does not work as expected. Calling `task` without a type argument will default to `Nothing`, I think, `anyTask` should be called instead.

silk-rules/src/main/scala/org/silkframework/rule/execution/CacheLoader.scala
```diff
@@ -63,7 +63,7 @@ class CacheLoader(source: DataSource,
   }
 
   private def retrieveEntities(implicit userContext: UserContext): CloseableIterator[Entity] = {
-    implicit val pluginContext: PluginContext = PluginContext(prefixes = prefixes, resources = EmptyResourceManager(), user = userContext)
+    implicit val pluginContext: PluginContext = PluginContext(prefixes = prefixes, resources = EmptyResourceManager(), user = userContext, taskResolver = TaskResolver.empty)
```
robertisele: Wouldn't this mean that the cache for link specs with rule blocks won't load? Maybe we should pipe the resolver through...

silk-rules/src/main/scala/org/silkframework/rule/RuleBlockSpec.scala
```diff
@@ -0,0 +1,282 @@
...
+        </OperatorTree>
+        {toXml[RuleLayout](value.layout)}
+        {toXml[UiAnnotations](value.uiAnnotations)}
+      </RuleBlockModel>
+    }
+  }
+}
+
+case class RuleBlockPort(id: Identifier = Operator.generateId,
+                         label: String = "",
```
robertisele: Isn't this a contradiction that the default for label is an empty string and thus an invalid value according to the assert? Maybe we should remove the default than.

silk-rules/src/main/scala/org/silkframework/rule/evaluation/statistics/LinkageRuleComplexity.scala
robertisele: It's minor, but we should probably count rule blocks as transformations? In that case, this could be changed to:

`transformationCount = ops.count(op => op.isInstanceOf[TransformInput] || op.isInstanceOf[RuleBlockInput])`

silk-core/src/main/resources/org/silkframework/LinkSpecificationLanguage.xsd
```diff
@@ -198,6 +224,111 @@
...
+                </xs:extension>
+            </xs:simpleContent>
+        </xs:complexType>
+    </xs:element>
+
+    <xs:element name="Port">
+        <xs:complexType>
+            <xs:sequence>
+                <xs:element ref="Description" minOccurs="0" maxOccurs="1"/>
+                <xs:element ref="ExampleValues" minOccurs="0" maxOccurs="1"/>
```
robertisele: Is the ExampleValues actually used? In RuleBlockPortXmlFormat, I don't see it.

silk-workbench/silk-workbench-rules/app/controllers/linking/LinkingTaskApi.scala
```diff
@@ -189,9 +190,9 @@ class LinkingTaskApi @Inject() (accessMonitor: WorkbenchAccessMonitor) extends I
   def putLinkSpec(projectName: String, taskName: String): Action[AnyContent] = RequestUserContextAction { request => implicit userContext =>
     val project = WorkspaceFactory().workspace.project(projectName)
     val task = project.task[LinkSpec](taskName)
-    implicit val prefixes = project.config.prefixes
-    implicit val resources = project.resources
-    implicit val readContext = ReadContext(resources, prefixes)
+    implicit val prefixes: Prefixes = project.config.prefixes
+    implicit val resources: ResourceManager = project.resources
+    implicit val readContext: ReadContext = ReadContext(resources, prefixes, taskResolver = TaskResolver.empty)
```
robertisele: This uses an empty TaskResolver, but the project needed to create one is just above.

silk-workbench/silk-workbench-rules/app/controllers/linking/LinkingTaskApi.scala
```diff
@@ -961,10 +1008,21 @@ class LinkingTaskApi @Inject() (accessMonitor: WorkbenchAccessMonitor) extends I
...
+                              "If true, changes the response to a JSON object with 'links' and 'ruleBlockInspection'.",
+                            required = false,
+                            in = ParameterIn.QUERY,
+                            schema = new Schema(implementation = classOf[Boolean], defaultValue = "false")
+                          )
+                          includeRuleBlockInspection: Boolean): Action[AnyContent] = RequestUserContextAction { implicit request => implicit userContext =>
     implicit val (project, task) = getProjectAndTask[LinkSpec](projectName, linkingTaskName)
     val sources = task.dataSources
-    implicit val readContext: ReadContext = ReadContext(prefixes = project.config.prefixes, resources = project.resources)
+    implicit val readContext: ReadContext = ReadContext(prefixes = project.config.prefixes, resources = project.resources,
```
robertisele, line 1025: Here it could also supply an actual task resolver

silk-core/src/main/resources/org/silkframework/LinkSpecificationLanguage.xsd
andreas-schultz: No, this is deprecated and was not removed. I will remove it.

silk-core/src/main/scala/org/silkframework/runtime/plugin/TaskResolver.scala
andreas-schultz: Right, this is not used, I remove it for now.

silk-rules/src/main/scala/org/silkframework/rule/evaluation/statistics/LinkageRuleComplexity.scala
andreas-schultz: Not sure, this code seems unused, we can also remove it. What do you think?

silk-rules/src/main/scala/org/silkframework/rule/RuleBlockSpec.scala
andreas-schultz: Good catch, this is fixed now.

silk-workspace/src/main/scala/org/silkframework/workspace/Project.scala
andreas-schultz: The validators here are only triggered when a task is created/updated. Removing a rule block that is in use, or more generically, any task that other tasks depend on, is already guarded by the backend.

J-x-D (changes requested):
submit empty review to enable comments (hopefully).

silk-rules/src/main/scala/org/silkframework/rule/RuleBlockSpec.scala
andreas-schultz: Yes, you mean `TaskValidationException`. I'll change that.

silk-rules/src/main/scala/org/silkframework/rule/RuleBlockSpec.scala
andreas-schultz: The UI currently prevents leaving the `displayOrder` empty, so this error only comes when the user has set at least 2 equal display orders.

andreas-schultz: I made the displayOrder required in the JSON serialization, so the direct API usage prevents not setting it.

silk-workspace/src/main/scala/org/silkframework/workspace/PluginUsage.scala
andreas-schultz: Right, I've added the case.

workspace/src/app/views/taskViews/linking/evaluation/tabView/LinkingEvaluationTabView.tsx
andreas-schultz: I changed the array to contain promises. But what do you mean by "It further fetches this data on every re-render"?

workspace/src/app/views/shared/RuleEditor/view/ruleNode/RuleNodeMenu.tsx
andreas-schultz: Have you tried clicking this menu item? It should give you a hint what you need to do then. Maybe you missed one of the main features of the rule block editor UI. ;)

silk-rules/src/main/scala/org/silkframework/rule/LinkSpec.scala
andreas-schultz: For me the doc reads that input and output tasks are only part of referenced tasks, but not exclusively.

silk-rules/src/main/scala/org/silkframework/rule/execution/CacheLoader.scala
andreas-schultz: Yes, that's what I had to do to fix another reported bugfurther above.

silk-rules/src/main/scala/org/silkframework/rule/RuleBlockSpec.scala
andreas-schultz: Yes, you are right. I've made the label and displayOrder consistent, also in the serializations.

silk-rules/src/main/scala/org/silkframework/rule/LinkSpec.scala
robertisele: But doesn't hurt to add it since that's maybe not obvious, no?

silk-rules/src/main/scala/org/silkframework/rule/evaluation/statistics/LinkageRuleComplexity.scala
robertisele: Yes, you are right, we can also just remove it. Would be easy to reconstruct if we ever need it again.

workspace/src/app/views/taskViews/linking/evaluation/tabView/LinkingEvaluationTabView.tsx
J-x-D: I mean: since it's not inside a useEffect, this code block will run every time the component renders.

silk-rules/src/main/scala/org/silkframework/rule/evaluation/statistics/LinkageRuleComplexity.scala
andreas-schultz: ok, I removed it.

silk-rules/src/main/scala/org/silkframework/rule/LinkSpec.scala
andreas-schultz: ok, I will improve the description.

PR #1044, opened 2026-05-11
CMEM-7596: Add per-value hash operator

edufuga:
## What and why

SPARQL 1.1 hash functions (`SHA256(?x)`, `MD5(?x)`, etc.) operate value by value — ten rows in, ten hashes out. Silk's existing "Input hash" operator combines all values from all ports into a single hash, which makes it impossible to replicate SPARQL hashing logic in a Transform rule without changing its semantics.

This PR closes that gap. The existing "Input hash" operator is renamed to make its combining semantics explicit, and a new "Per-value hash" operator is added that hashes each value independently — one hash per value, cardinality preserved. Both operators share a common `HashTransformer` trait that provides the hex encoding and digest logic.

## Changes

The existing "Input hash" operator has been renamed to "Combined input hash". Its label and description now make the combining behavior unambiguous at a glance. The plugin ID (`inputHash`) is unchanged so that existing serialized configurations continue to work without migration. Its documentation has been rewritten from scratch and its test suite expanded from one case to eleven, covering multiple values, order sensitivity, multi-port combination, and all five SPARQL 1.1 algorithms.

The new "Per-value hash" operator (`perValueHash`) hashes each input value independently, producing one hash per value in the same order as the input. It accepts exactly one input port and throws `IllegalArgumentException` if more are connected. It uses the same open-ended JVM MessageDigest algorithm list as the existing operator, covering the full SPARQL 1.1 set and more. Its documentation and test suite (12 cases) are new.

A `HashTransformer` trait has been extracted to hold the shared hex encoding and digest logic; both plugins implement only their `apply` method and inherit everything else. The `HashAlgorithmAutoCompletionProvider`, previously a nested class inside `InputHashTransformer.scala`, has been moved to its own file and is shared by both operators. A `relatedPlugins` annotation on each operator links to the other, making the distinction discoverable in the UI.

## Tests

`InputHashTransformerTest` (existing, expanded) — 11 cases: single value, multiple values, order sensitivity, multi-port combination, all five SPARQL 1.1 algorithms (MD5, SHA-1, SHA-256, SHA-384, SHA-512), empty input.

`PerValueHashTransformerTest` (new) — 12 cases: single value, multiple values, all five algorithms, empty input, single-input constraint violation (`IllegalArgumentException`), empty algorithm string (`IllegalArgumentException`), invalid algorithm name (`NoSuchAlgorithmException`).

## Acceptance criteria

- "Input hash" renamed to "Combined input hash" — plugin ID unchanged for backward compatibility
- New "Per-value hash" operator hashes each value independently and enforces a single input source
- Both operators support MD5, SHA-1, SHA-256, SHA-384, SHA-512 and the full JVM MessageDigest provider set

### DIFF
--- a/silk-rules/src/main/scala/org/silkframework/rule/plugins/transformer/value/HashTransformer.scala
+++ b/silk-rules/src/main/scala/org/silkframework/rule/plugins/transformer/value/HashTransformer.scala
@@ -16,18 +16,18 @@ trait HashTransformer extends Transformer {
   /** The hash algorithm name, passed directly to [[java.security.MessageDigest#getInstance]].
    *  Must be recognized by the JVM's security provider registry; unrecognized names cause
    *  [[java.security.NoSuchAlgorithmException]] at runtime. */
-  protected def algorithm: String
+  def algorithm: String
 
   /** Encodes a byte array as a fixed-length lowercase hex string. Each byte maps to exactly two
    *  characters; without zero-padding, a byte with value 5 would render as "5" instead of "05",
    *  producing a variable-length string that breaks any downstream comparison or length check. */
-  protected def toHex(bytes: Array[Byte]): String =
+  def toHex(bytes: Array[Byte]): String =
     bytes.map("%02x".format(_)).mkString
 
   /** Hashes a single string value and returns its hex digest. A fresh [[java.security.MessageDigest]]
    *  instance is created on every call: [[java.security.MessageDigest]] is stateful and not thread-safe,
    *  so reusing one instance across concurrent calls would corrupt the output. */
-  protected def hashValue(v: String): String = {
+  def hashValue(v: String): String = {
     val digest = MessageDigest.getInstance(algorithm)
     digest.update(v.getBytes(StandardCharsets.UTF_8))
     toHex(digest.digest())
--- a/silk-rules/src/main/scala/org/silkframework/rule/plugins/transformer/value/HashTransformer.scala
+++ b/silk-rules/src/main/scala/org/silkframework/rule/plugins/transformer/value/HashTransformer.scala
@@ -9,9 +9,13 @@ import java.util.HexFormat
 /**
  * Digest operations for hash transformer plugins.
  *
- * Provides [[toHex]] and [[hashValue]] as concrete implementations over the abstract [[algorithm]]
- * parameter, which each concrete plugin supplies via its own plugin parameter. Subclasses implement
- * [[apply]] according to their combining or per-value semantics.
+ * Provides [[toHex]], [[hashValue]], and [[withDigest]] as concrete implementations over the abstract
+ * [[algorithm]] parameter, which each concrete plugin supplies via its own plugin parameter. Subclasses
+ * implement [[apply]] according to their combining or per-value semantics:
+ *  - [[hashValue]] is suited to per-value operators: it is a complete, self-contained operation on a
+ *    single string.
+ *  - [[withDigest]] is suited to combining operators: it loans a reset digest for the duration of a
+ *    single lambda, leaving the caller responsible for updates and finalization within that scope.
  */
 trait HashTransformer extends Transformer {
   /** The hash algorithm name, passed directly to [[java.security.MessageDigest#getInstance]].
@@ -24,7 +28,7 @@ trait HashTransformer extends Transformer {
   @transient private lazy val hexFormat: HexFormat = HexFormat.of()
 
   // @transient: not serialized. lazy: recreated on first use after deserialization.
-  @transient private lazy val localDigest: ThreadLocal[MessageDigest] =
+  @transient private lazy val threadLocalDigest: ThreadLocal[MessageDigest] =
     ThreadLocal.withInitial(() => MessageDigest.getInstance(algorithm))
 
   /** Encodes a byte array as a fixed-length lowercase hex string. Each byte maps to exactly two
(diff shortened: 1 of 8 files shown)


silk-rules/src/main/scala/org/silkframework/rule/plugins/transformer/value/HashTransformer.scala
```diff
@@ -0,0 +1,35 @@
...
+   *  characters; without zero-padding, a byte with value 5 would render as "5" instead of "05",
+   *  producing a variable-length string that breaks any downstream comparison or length check. */
+  def toHex(bytes: Array[Byte]): String =
+    bytes.map("%02x".format(_)).mkString
+
+  /** Hashes a single string value and returns its hex digest. A fresh [[java.security.MessageDigest]]
+   *  instance is created on every call: [[java.security.MessageDigest]] is stateful and not thread-safe,
+   *  so reusing one instance across concurrent calls would corrupt the output. */
+  def hashValue(v: String): String = {
+    val digest = MessageDigest.getInstance(algorithm)
```
robertisele: Might be a bit inefficient to create a new instance for each value, but I suppose using a ThreadLocal instance is overkill, so we probably just need to be aware that we might have to improve this in the future.

silk-rules/src/main/scala/org/silkframework/rule/plugins/transformer/value/InputHashTransformer.scala
```diff
@@ -1,77 +1,103 @@
...
   require(algorithm.trim.nonEmpty, "Algorithm must not be empty. Please specify an algorithm, such as 'SHA256'.")
 
   override def apply(values: Seq[Seq[String]]): Seq[String] = {
-    val hashSum = MessageDigest.getInstance(algorithm)
-    for(value <- values; v <- value) {
-      hashSum.update(v.getBytes(StandardCharsets.UTF_8))
-    }
-    // Convert the byte array to a hexadecimal string
-    Seq(hashSum.digest().map("%02x".format(_)).mkString)
+    val digest = MessageDigest.getInstance(algorithm)
```
robertisele: Could this also call the hashValue function in the parent trait to avoid code duplication?

silk-rules/src/main/scala/org/silkframework/rule/plugins/transformer/value/HashTransformer.scala
```diff
@@ -0,0 +1,35 @@
...
+  /** The hash algorithm name, passed directly to [[java.security.MessageDigest#getInstance]].
+   *  Must be recognized by the JVM's security provider registry; unrecognized names cause
+   *  [[java.security.NoSuchAlgorithmException]] at runtime. */
+  def algorithm: String
+
+  /** Encodes a byte array as a fixed-length lowercase hex string. Each byte maps to exactly two
+   *  characters; without zero-padding, a byte with value 5 would render as "5" instead of "05",
+   *  producing a variable-length string that breaks any downstream comparison or length check. */
+  def toHex(bytes: Array[Byte]): String =
+    bytes.map("%02x".format(_)).mkString
```
robertisele: Wouldn't it be slightly better and more efficient to use a stored   java.util.HexFormat instance here?

silk-rules/src/main/scala/org/silkframework/rule/plugins/transformer/value/HashTransformer.scala
edufuga: The overhead is a real concern, the allocation adds up when hashing large numbers of values, and a fresh instance per call is wasteful. A thread-local instance is the right pattern: each thread keeps its own, reused across calls without any thread-safety risk. I went ahead and added it.

silk-rules/src/main/scala/org/silkframework/rule/plugins/transformer/value/InputHashTransformer.scala
edufuga: It's not actually a duplication. The two operations are structurally different: hashValue is a complete self-contained operation on a single value, while the combining operator needs to keep a single digest open across all values and all ports before finalizing once. Delegating to hashValue would break that, since it finalizes immediately.

This is actually why the trait has the structure it has: hashValue exists for the per-value operator, where the per-value semantics map onto it cleanly. The combining operator implements its own apply against the shared lower-level primitives, because its logic can't be expressed through hashValue.

silk-rules/src/main/scala/org/silkframework/rule/plugins/transformer/value/HashTransformer.scala
edufuga: A stored instance is indeed cleaner and avoids the per-byte format string allocation. I switched to HexFormat.

silk-rules/src/main/scala/org/silkframework/rule/plugins/transformer/value/InputHashTransformer.scala
robertisele: Yes, you are right! We could add a new hashValueS function to the trait though to benefit from the ThreadLocal performance improvements?

silk-rules/src/main/scala/org/silkframework/rule/plugins/transformer/value/HashTransformer.scala
robertisele: great

silk-rules/src/main/scala/org/silkframework/rule/plugins/transformer/value/HashTransformer.scala
robertisele: Cool that should be way faster 👍

silk-rules/src/main/scala/org/silkframework/rule/plugins/transformer/value/InputHashTransformer.scala
edufuga: Good point on the performance side. Rather than adding combining semantics to the trait, I added `withDigest` — a loan pattern over the thread-local digest — and `updateWith` for the encoding. Hashing over several values is a subtype concern, not a trait concern.

Both operators now go through the same trait methods. `InputHashTransformer.apply` uses `withDigest` with a local `updateAll` over `values.flatten` instead of allocating a fresh digest on every call.

robertisele: makes sense, good change.

robertisele (approved):

PR #1045, opened 2026-05-12
Workflow execution variables

robertisele:


silk-core/src/main/resources/org/silkframework/plugins/operations/SetExecutionVariableOperator.md
```diff
@@ -0,0 +1,25 @@
+Sets a single **execution-scope** template variable from this operator's input and passes the input through
+unchanged, so it can be inserted anywhere in a workflow chain.
+
+The variable is written to the `execution` scope and can be read by any downstream node as
```
andreas-schultz: This paragraph is a little bit confusing because it mixes two different concerns. `referencing an execution variable that has not been set fails` should in my opinion moved into a different paragraph, since it is about the read aspect of variables, the other sentences are about writing those variables.

silk-core/src/main/scala/org/silkframework/runtime/templating/TemplateVariablesManager.scala
```diff
@@ -1,12 +1,39 @@
...
+    * Returns the parent scope variables that can be referenced in templates.
+    * For project variables, this returns the global variables.
+    * For the execution variables of a task, this returns the global and project variables.
+    */
+  def parentVariables: TemplateVariables
+
+  /**
+    * The scope that all variables managed by this manager must be in.
+    */
+  def variableScope: Seq[String]
```
andreas-schultz: Maybe replace Seq[String] by a more specific case class? This generic data structure is now used in multiple places.

silk-core/src/main/scala/org/silkframework/plugins/operations/SetExecutionVariableOperator.scala
```diff
@@ -0,0 +1,47 @@
...
+case class SetExecutionVariableOperator(@Param("Name of the execution variable to set. " +
+                                               "It is written to the 'execution' scope and addressed downstream as 'execution.<name>'.")
+                                        variableName: String = "myVariable",
+                                        @Param(label = "Source path",
+                                          value = "Optional path/attribute of the input that supplies the value. " +
+                                            "If left empty, the first value of the input is used.")
+                                        sourcePath: String = "") extends CustomTask {
+
+  /** A single input of any schema, so the operator can be placed after any node. */
+  override def inputPorts: InputPorts = FixedNumberOfInputs(Seq(UnknownSchemaPort))
```
andreas-schultz: If sourcePath is defined, shouldn't this be a fixed schema? Else a dataset cannot be the input.

silk-workbench/silk-workbench-core/app/controllers/workspaceApi/coreApi/VariableTemplateApi.scala
```diff
@@ -56,12 +57,21 @@ class VariableTemplateApi @Inject()() extends InjectedController with UserContex
...
+                     name = "task",
+                     description = "The task identifier. If provided, retrieves the execution variables of that task instead of project variables.",
+                     required = false,
+                     in = ParameterIn.QUERY,
+                     schema = new Schema(implementation = classOf[String])
+                   )
+                   task: Option[String]): Action[AnyContent] = RequestUserContextAction { implicit request => implicit userContext =>
     val project = WorkspaceFactory().workspace.project(projectName)
-    val allVariables = project.templateVariables.all
+    val manager = project.variablesManager(task)
```
andreas-schultz: This currently returns only the variables defined on that task. For e.g. workflows it would also be interesting to know all defined execution variables of sub-tasks. Either by default or with a transitive flag.

silk-workbench/silk-workbench-workflow/app/controllers/workflowApi/variableWorkflow/VariableWorkflowRequestUtils.scala
```diff
@@ -304,28 +310,34 @@ object VariableWorkflowRequestUtils {
...
       configParameters = ParameterValues(Map(
         "configuration" -> ParameterStringValue(workflowConfig.toString()),
         "configurationType" -> ParameterStringValue(jsonMimeType),
-        "optionalPrimaryResourceManager" -> ParameterObjectValue(OptionalPrimaryResourceManagerParameter(Some(variableWorkflowFileResourceManager)))
+        "optionalPrimaryResourceManager" -> ParameterObjectValue(OptionalPrimaryResourceManagerParameter(Some(variableWorkflowFileResourceManager))),
+        WorkflowExecutorFactory.EXECUTION_VARIABLES_PARAMETER -> ParameterObjectValue(TemplateVariablesParameter(executionVars))
       )),
-      variableDataSinkConfig = replaceableDataSinkConfigOpt.map(_.mimeType)
+      variableDataSinkConfig = replaceableDataSinkConfigOpt.map(_.mimeType),
+      executionVariables = executionVars
```
andreas-schultz, line 324: This will only work if the request is a JSON request. For other data types, form-urlencoded or multipart requests this won't work.

silk-core/src/main/resources/org/silkframework/plugins/operations/SetExecutionVariableOperator.md
robertisele: I improved this doc and restructured the paragraphs.

silk-core/src/main/scala/org/silkframework/plugins/operations/SetExecutionVariableOperator.scala
robertisele: Yes, right, I changed this to a fixed schema, if the path is defined.

silk-core/src/main/scala/org/silkframework/runtime/templating/TemplateVariablesManager.scala
robertisele: Make sense! I introduced a new VariableScope case class and replaced all usages. This is a large commit, but I think for the better.

silk-workbench/silk-workbench-workflow/app/controllers/workflowApi/variableWorkflow/VariableWorkflowRequestUtils.scala
robertisele: I have added query parameters of the form `variable-*` to cover this.

silk-workbench/silk-workbench-core/app/controllers/workspaceApi/coreApi/VariableTemplateApi.scala
robertisele: I have added a "transitive" flag. I think we need the flag because this endpoint is also used by the "Execution Variables" widget and this should only receive the direct variables.

andreas-schultz: It now supports sub-workflows, but what about transform/linking tasks?
When the workflow does not duplicate these variables, they are still needed for the execution and will lead to a runtime error.

andreas-schultz: moved to follow-up

PR #1046, opened 2026-05-18
Maintain GUI elements libraries (CMEM-7611)

haschek:


workspace/config/webpack.di.config.js
```diff
@@ -253,6 +253,7 @@ module.exports = function (webpackEnv, isWatch) {
                 "#minpath": "vfile/lib/minpath.browser.js",
                 "#minproc": "vfile/lib/minproc.browser.js",
                 "#minurl": "vfile/lib/minurl.browser.js",
+                "es-toolkit/compat": "es-toolkit/compat.d.ts",
```
J-x-D: is that correct? Why do we need the declaration file here?

package.json
```diff
@@ -27,6 +27,7 @@
     "**/semver-regex": "3.1.4",
     "**/@types/react": "^18.2.0",
     "**/url-parse": "1.5.9",
-    "**/react-redux": "^9.1.2"
+    "**/react-redux": "^9.1.2",
+    "**/react-syntax-highlighter": "^15.0.0"
```
J-x-D, line 31: in gui-elements you defined version "^16.1.1" - can't you use that here as well?

workspace/config/webpack.di.config.js
haschek: Good find, this was wrong. Should be the correct target now.

package.json
haschek: Yes, v16 would work, but webpack4 in Silk cannot resolve the export syntax of the `package.json` from the package. We could add a lot of (a lot of!) webpack4 aliases to the config, but it was way easier to downgrade the package to v15 via version resolution.

PR #1050, opened 2026-06-02
Add JSON to File operator (CMEM-7616)

edufuga:
_This PR is based on develop. It depends on PR [#1041] (CMEM-7427)._

CMEM-7616: JSON to File operator
=================================

Adds an operator that reads a JSON string from a field on each input entity and writes it to a file, surfacing the result as a `FileEntity` for downstream file-backed operators or datasets.

Why
---

Parse JSON consumes a JSON string and produces structured entities; it does not preserve the string. Users who need to pass the JSON value to a file-backed operator downstream — for archival, transfer, or further processing — had no operator for that. `JsonToFileOperator` fills that gap.

Changes
------

`JsonToFileOperator` is a `CustomTask` registered in `JsonPlugins`. Its input port mirrors Parse JSON: one entity carrying a JSON string. Its output is a `FileEntity`. For each input entity the operator reads the value at the configured input path, validates it as JSON, and writes it to a file tagged `application/json`. When the input contains more than one entity, a zero-based index suffix is appended to the filename before the extension to keep output names unique and predictable. If no input path is configured, the operator defaults to the first field on the input entity, consistent with Parse JSON's behaviour.

An optional `zipOutput` flag switches to ZIP mode: instead of one file per entity, all entities are packed into a single ZIP archive with one entry per entity. Entry names follow the same suffix scheme as the single-file path. The produced `FileEntity` carries `application/zip` in ZIP mode. If the write loop fails mid-archive the partial temp file is deleted before the exception propagates, so callers never receive a broken ZIP.

An optional `outputProperty` parameter wraps the JSON value in an object under the given key before writing. With `outputProperty` set to `payload`, an input value `{"name":"Alice"}` is written as `{"payload":{"name":"Alice"}}`. When left empty — the default — the value is written as-is. The wrapping applies in both normal and ZIP modes. Internally, `validateJson` is replaced by `parseJson` returning a `JsonNode`, which validates and supplies the parsed node for wrapping in a single pass.

A third `outputMode`, `jsonArray`, merges all input entities' values into a single JSON array file, splicing each element verbatim.

`JsonParserTask` gains a mutual `relatedPlugins` entry pointing at `JsonToFileOperator`, and vice versa, so the workbench surfaces the relationship from either side.

Tests
-----

`LocalJsonToFileOperatorExecutorTest` covers the three output modes independently. For `file` mode: single and multi-entity output, the literal-versus-suffixed filename rule, `application/json` MIME-type propagation, and input-path defaulting. For `zip` mode: single-entity packing under `entry.json`, multi-entity packing with suffixed entry names, a named ZIP container, a single literal entry name, an explicitly configured MIME type, empty input, and a 500-entity scale run confirming the write loop and entry naming hold at volume. For `jsonArray` mode: single and multi-entity merging into one array in input order, a single named output file with no index suffix, the default and explicitly configured MIME types, and the value-fidelity cases below.

Invalid input is skipped per entity rather than aborting the task, asserted in every mode: an empty value and an unparseable value are each skipped and recorded as a warning on the execution report while the valid entities still produce output; a mixed batch writes the valid records and skips the bad one, with contiguous file and ZIP entry names across the gap; a fully-skipped run yields the mode's natural empty output — no files, an empty `[]`, or a ZIP with no entries; and the report's `entityCount` (plus the skip-warning count) is asserted in `file`, `zip`, and `jsonArray` modes, with an all-valid merged run asserting no warnings.

Byte fidelity in `jsonArray` mode is pinned directly: trailing decimal zeros, large-magnitude integers, plain and multiple-of-ten integers, scalar / array / null elements, duplicate object keys, and pretty-printed internal formatting are written through verbatim; a non-JSON token after a valid value and a comma-separated trailing value at the root are rejected (skipped and warned); and whitespace padding around an element is preserved. `outputProperty` wrapping is covered across modes — single and multiple entities, ZIP entries, a JSON-array input, a skipped invalid value, and number- and whitespace-preservation in `file` and `zip` when the wrapper is set.

`LocalJsonToFileWorkflowTest` runs the operator end-to-end in a three-node in-memory workflow: `file` mode asserting the output resource holds the source JSON byte-for-byte, `zip` mode asserting entry names and content across two entities, `jsonArray` mode asserting the merged array, and a skip case asserting an invalid record is dropped while the valid ones flow through to the downstream dataset.

---

Task: https://jira.eccenca.com/browse/CMEM-7616


silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/JsonToFileOperator.scala
```diff
@@ -0,0 +1,71 @@
...
+
+@Plugin(
+  id = JsonToFileOperator.pluginId,
+  label = "JSON to File",
+  description = "Writes a JSON string held in a field on each valid incoming entity to a file. Depending on the output mode, it produces one file per entity, packs all entities into a single ZIP archive, or merges them into a single JSON array file. Produces a file entity downstream, suitable for wiring into a file-backed dataset or any operator that consumes file entities.",
+  documentationFile = "JsonToFileTaskDocumentation.md",
+  relatedPlugins = Array(
+    new PluginReference(
+      id = JsonParserTask.pluginId,
+      description = "JSON to File writes the JSON string on each input entity to a file; Parse JSON parses the same kind of input into structured entities driven by a downstream schema."
```
andreas-schultz: Maybe change "on each input entity" -> "from each..."

silk-core/src/main/scala/org/silkframework/execution/typed/FileEntitySchema.scala
```diff
@@ -99,13 +99,22 @@ object FileEntity {
...
+   * @param suffix   Suffix (extension) for the generated filename. Ignored when `name` is provided.
+   * @param name     If provided, the file is created with this exact filename in the system temp directory.
+   *                 If absent, a unique name is generated from `prefix` and `suffix`.
+   * @param mimeType Optional MIME type to attach to the produced FileEntity.
    */
-  def createTemp(prefix: String, suffix: String = ".tmp"): FileEntity = {
-    val tempFile = File.createTempFile(prefix, suffix)
+  def createTemp(prefix: String, suffix: String = ".tmp", name: Option[String] = None, mimeType: Option[String] = None): FileEntity = {
+    val tempFile = name match {
+      case Some(n) => new File(System.getProperty("java.io.tmpdir"), n)
```
andreas-schultz: This new approach of creating temp files inrtoduces a security bug. It does not check that the file is actually inside the temp directory. E.g. "../" will escape it.
Even if this would be prevented, this would allow to access any temp file the process has read or write access to.
Also, regarding the "JSON to file" usage, multiple executions with the same output file parameter will try to write into the exact same file.

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/LocalJsonToFileOperatorExecutor.scala
```diff
@@ -0,0 +1,235 @@
...
+    * matching the byte preservation of file and zip modes. Entities whose value is not valid JSON are skipped and
+    * recorded as warnings on the execution report. */
+  private def executeMergedJson(task: Task[JsonToFileOperator],
+                               entities: IndexedSeq[Entity],
+                               pathIndex: Int,
+                               trimmedFileName: String,
+                               context: ActivityContext[ExecutionReport])
+                              (implicit pluginContext: PluginContext): Option[LocalEntities] = {
+    if (entities.isEmpty) {
+      return Some(FileEntitySchema.create(Seq.empty, task))
```
andreas-schultz: This currently will not write a file with the empty array content `[]`, but no file. If the output is e.g. a JSON dataset, the previous, stale file is kept, which is probably unexpected.

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/LocalJsonToFileOperatorExecutor.scala
```diff
@@ -0,0 +1,235 @@
...
+  /** Executes ZIP mode: writes all valid entities as entries into a single ZIP file and returns one FileEntity.
+    * Entities whose value is not valid JSON are skipped and recorded as warnings on the execution report. */
+  private def executeZip(task: Task[JsonToFileOperator],
+                         entities: IndexedSeq[Entity],
+                         pathIndex: Int,
+                         trimmedFileName: String,
+                         context: ActivityContext[ExecutionReport])
+                        (implicit pluginContext: PluginContext): Option[LocalEntities] = {
+    if (entities.isEmpty) {
+      return Some(FileEntitySchema.create(Seq.empty, task))
```
andreas-schultz: Same as with the JSON array output, this will not write an empty zip, which might leave stale files in a JSON output dataset.

silk-core/src/main/scala/org/silkframework/execution/typed/FileEntitySchema.scala
andreas-schultz: If the file name should be "preserved", this should probably be done with a different approach.

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/LocalJsonToFileOperatorExecutor.scala
```diff
@@ -0,0 +1,235 @@
...
+                       context: ActivityContext[ExecutionReport] = new ActivityMonitor(getClass.getSimpleName))
+                      (implicit pluginContext: PluginContext): Option[LocalEntities] = {
+    if (inputs.size != 1) throw TaskException("'JSON to File' takes exactly one input!")
+    val entityTable = inputs.head
+    val pathIndex = task.data.parsedInputPath match {
+      case Some(path) => entityTable.entitySchema.indexOfPath(path)
+      case None => 0 // Take the value of the first path
+    }
+    val trimmedFileName = task.data.outputFileName.trim
+    val entities = entityTable.entities.toIndexedSeq
```
andreas-schultz: This will materialize all entities into memory. Most tasks if possible output streams in order to keep the memory-footprint low.

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/LocalJsonToFileOperatorExecutor.scala
```diff
@@ -0,0 +1,235 @@
...
+      FileEntity.createTemp(TempFilePrefix, ".zip", mimeType = Some(effectiveMimeType))
+    } else {
+      allocateNamedFileEntity(trimmedFileName, 0, multiple = false, Some(effectiveMimeType))
+    }
+    try {
+      zipFileEntity.file.write() { outputStream =>
+        val zip = new ZipOutputStream(outputStream)
+        for ((content, index) <- contents.zipWithIndex) {
+          val entryName = if (trimmedFileName.nonEmpty) {
+            if (multiple) suffixedName(trimmedFileName, index) else trimmedFileName
```
andreas-schultz: Similar to the temp file name creation, ZIP entries should probably also be sanitized.

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/LocalJsonToFileOperatorExecutor.scala
```diff
@@ -0,0 +1,235 @@
...
+  /** Executes file mode: writes one file per valid entity and returns one FileEntity per valid entity. Entities whose
+    * value is not valid JSON are skipped and recorded as warnings on the execution report. */
+  private def executeFile(task: Task[JsonToFileOperator],
+                          entities: IndexedSeq[Entity],
+                          pathIndex: Int,
+                          trimmedFileName: String,
+                          context: ActivityContext[ExecutionReport])
+                         (implicit pluginContext: PluginContext): Option[LocalEntities] = {
+    if (entities.isEmpty) {
+      return Some(FileEntitySchema.create(Seq.empty, task))
```
andreas-schultz: This early return prevents that the execution report is ever updated.

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/JsonToFileOperator.scala
```diff
@@ -0,0 +1,71 @@
...
+    )
+  )
+)
+case class JsonToFileOperator(@Param("The Silk path expression of the input entity that contains the JSON string. " +
+                                  "If not set, the value of the first defined property will be taken.")
```
andreas-schultz: `first defined property` might be misleading. "defined" could be misunderstood as the first property that has a non-empty value.

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/JsonToFileOperator.scala
```diff
@@ -0,0 +1,71 @@
...
+                              inputPath: String = "",
+                              @Param("Filename for the produced file. If left empty, an auto-generated temporary name is used. " +
+                                  "In file output mode, when the input contains more than one entity, an index suffix is appended " +
+                                  "before the extension (e.g. `out-0.json`, `out-1.json`) to keep filenames unique. " +
```
andreas-schultz: The description and the code assumes that a file name containing colons have a file suffix. E.g. something like ".hiddenfile" would end up getting an indexed name like "-0.hiddenfile". Not sure what a good heuristic would be though, maybe at least not adding the index part to the start.

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/JsonToFileOperator.scala
```diff
@@ -0,0 +1,71 @@
...
+                                  "In ZIP output mode, this name is used for the ZIP container file and as the base for entry names inside the archive. " +
+                                  "In jsonArray output mode there is always a single output file, so the name is used as-is with no index suffix.")
+                              outputFileName: String = "",
+                              @Param("MIME type of the produced file.")
+                              mimeType: String = "application/json",
+                              @Param("Output mode: file writes one file per entity, zip packs all entities into a single ZIP archive, " +
```
andreas-schultz: The user will only see the display names of the enums and not the IDs, so this should mention the display names instead.

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/LocalJsonToFileOperatorExecutor.scala
```diff
@@ -0,0 +1,235 @@
...
+                               trimmedFileName: String,
+                               context: ActivityContext[ExecutionReport])
+                              (implicit pluginContext: PluginContext): Option[LocalEntities] = {
+    if (entities.isEmpty) {
+      return Some(FileEntitySchema.create(Seq.empty, task))
+    }
+    val outputMimeType = Some(task.data.mimeType)
+    val outputProperty = task.data.outputProperty
+    val (contents, warnings) = partitionValid(entities)(rawContent(_, pathIndex, outputProperty))
+    val serialized = contents.mkString("[", ",", "]")
```
andreas-schultz: The entities are currently already loaded into memory. But when this changes, creating the array string in-memory would probably not be a good idea.

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/JsonToFileOperator.scala
edufuga: Done — changed to "from each input entity."

silk-core/src/main/scala/org/silkframework/execution/typed/FileEntitySchema.scala
edufuga: You're right, and I don't think this is worth trying to fix in place. I removed the `name` parameter from `createTemp` entirely — it's back to the original `createTemp(prefix, suffix)` form. That's collision-safe and non-traversable by construction, so there's no path-joining logic left for `../` to exploit.

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/LocalJsonToFileOperatorExecutor.scala
edufuga: Fixed.

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/LocalJsonToFileOperatorExecutor.scala
edufuga: Fixed.

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/LocalJsonToFileOperatorExecutor.scala
edufuga: Fixed. Removed the `toIndexedSeq` call and the `partitionValid` helper that depended on it. Entities are now consumed one at a time from the iterator via a shared `processSkippingInvalid` helper that each mode drives with its own per-entity callback — at most one entity's content is in memory at any point.

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/LocalJsonToFileOperatorExecutor.scala
edufuga: Fixed (see second comment).

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/LocalJsonToFileOperatorExecutor.scala
edufuga: Fixed; same early-return removal as comments 3 and 4.

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/JsonToFileOperator.scala
edufuga: Indeed. Fixed.

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/JsonToFileOperator.scala
edufuga: I removed `outputFileName` and the whole suffix-insertion mechanism as part of the fix for comment 5, so there's no dotfile heuristic left to get wrong.

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/JsonToFileOperator.scala
edufuga: Fixed — the description now uses the actual display names ("One file per entity", "ZIP archive", "Merged JSON array") instead of the internal IDs (`file`, `zip`, `jsonArray`).

silk-plugins/silk-plugins-json/src/main/scala/org/silkframework/plugins/dataset/json/LocalJsonToFileOperatorExecutor.scala
edufuga: Fixed. Nothing accumulates the full array in memory anymore.

PR #1053, opened 2026-06-10
Refactor workbench config

andreas-schultz:


silk-workbench/silk-workbench-core/app/config/WorkbenchConfig.scala
```diff
@@ -81,6 +82,18 @@ object WorkbenchConfig {
     }
   }
 
+  def configuredPublicBaseUrl(config: TypesafeConfig): String = {
+    val protocol = if(useHttps(config)) "https" else "http"
+    val publicHost = host(config).getOrElse("localhost:9000")
```
robertisele: Is it intentional that the fallback neither includes the configured port nor the `applicationContext`? Otherwise `dataIntegrationUrl `could be used instead (but needs to be fixed to include https).

andreas-schultz: I've fixed this and refactored the overall code to make it more clear.

PR #1056, opened 2026-06-15
Remove `DatasetAccess` from `Dataset`, access datasets through their executor

robertisele:


silk-core/src/main/scala/org/silkframework/execution/local/LocalDatasetExecutor.scala
andreas-schultz: Here the sqlEndpoint is still directly accessed, which might not be a problem, but looks inconsistent.

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/datasets/InMemoryDataset.scala
```diff
@@ -91,22 +89,7 @@ case class InMemoryDataset(
...
-
-  override def entitySink(implicit userContext: UserContext): EntitySink =
-    new SparqlSink(SparqlParams(), sparqlEndpoint, dropGraphOnClear = dropGraph)
-
-  override def linkSink(implicit userContext: UserContext): LinkSink =
-    new SparqlSink(SparqlParams(), sparqlEndpoint, dropGraphOnClear = dropGraph)
-
-  override def tripleSink(implicit userContext: UserContext): TripleSink =
-    new SparqlSink(SparqlParams(), sparqlEndpoint, dropGraphOnClear = dropGraph)
+  def sparqlEndpoint: SparqlEndpoint = endpoint
```
andreas-schultz: This still allows access without executor. Currently it is only used in tests and benchmarks. Maybe this could be restricted or at least documented that it is not for production use?

silk-core/src/main/scala/org/silkframework/execution/local/LocalDatasetExecutor.scala
robertisele: I moved the sqlEndpoint to a new `SqlDatasetAccess `class (same pattern as for the `RdfDatasetAccess`).

silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/datasets/InMemoryDataset.scala
robertisele: I removed this here and updated the call sites.

andreas-schultz (approved):

PR #1063, opened 2026-06-26
Workflow port schema extension

andreas-schultz:


silk-workbench/silk-workbench-workspace/app/controllers/workspaceApi/WorkspaceVocabularyApi.scala
```diff
@@ -125,4 +173,106 @@ class WorkspaceVocabularyApi extends InjectedController with UserContextActions
...
+    val shortened = prefixes.shorten(absoluteUri)
+    Option.when(shortened != absoluteUri)(shortened)
+  }
+
+  private def enrichLabels(results: Seq[VocabularyLookupResult],
+                           preferredLanguage: Option[String])
+                          (implicit userContext: UserContext): Seq[VocabularyLookupResult] = {
+    val languageOpt = preferredLanguage
+      .map(_.trim)
+      .filter(language => language.nonEmpty && !language.equalsIgnoreCase("en"))
```
robertisele, line 243: Shouldn't this also catch compound English language strings, such as "en-US"?

silk-workbench/silk-workbench-workspace/app/controllers/workspaceApi/WorkspaceVocabularyApi.scala
```diff
@@ -125,4 +173,106 @@ class WorkspaceVocabularyApi extends InjectedController with UserContextActions
...
+      uri = absoluteUri,
+      kind = Some(kind),
+      label = info.label.orElse(info.altLabels.headOption),
+      description = info.description,
+      prefixedUri = prefixedUri(absoluteUri),
+      graphUri = info.vocabularyUri
+    )
+  }
+
+  private def resolveVocabularyEntry(absoluteUri: String,
```
robertisele, line 221: Wouldn't this get slow for large vocabularies? It looks like this is called for each property separately. 

The `Vocabulary` class currently scans through all classes and properties, so we might have to add indices for both in there in the future...

andreas-schultz: Yes, you are right, I assumed (incorrectly) that the class/property operations are O(1).

silk-workbench/silk-workbench-workspace/app/controllers/workspaceApi/WorkspaceVocabularyApi.scala
andreas-schultz: Not sure, this should mostly handle the use case of the UI language, which is set to "en" in combination with the vocabulary cache which already prefers "en" values. Maybe for these compound language strings you want to have the resolution instead.

silk-workbench/silk-workbench-workspace/app/controllers/workspaceApi/WorkspaceVocabularyApi.scala
andreas-schultz: I've added Map based lookup.

silk-workbench/silk-workbench-workspace/app/controllers/workspaceApi/WorkspaceVocabularyApi.scala
robertisele: Ok, I didn't know that this will always be just "en", I thought it might be coming from a browser, which also could send "en-US"

PR #1066, opened 2026-06-29
Support copy-to-clipoard for entity/graph URIs

andreas-schultz:
Also:

- Make transform/linking evaluation view values selectable for copy/paste


workspace/src/app/hooks/useCopyButton.tsx
```diff
@@ -4,7 +4,7 @@ import { useTranslation } from "react-i18next";
 import { copyToClipboard } from "../utils/copyToClipboard";
 import useErrorHandler from "./useErrorHandler";
 
-interface ICopyData {
+export interface ICopyData {
```
haschek: Would it make sense to rename the interface to our currently used naming scheme `ComponentNameProps` before we start to export it?

workspace/src/app/views/shared/TaskConfig/TaskConfigPreview.tsx
haschek, line 115: 1. As it is only one property it could have set directly, not via variable.
2. More important: This looks like a fix that should be applied to the GUI component directly. (the `line-height: 1` originates from a global reset style.

I recognized that this code piece is already 6 years old but the display and style of the property value was adjusted in this branch again.

At least we should create  ticket, referencing this workaround.

workspace/src/app/views/shared/CompactCopyUriButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+                    data-test-id={dataTestId}
+                    name={isCopied ? "state-confirmed" : "item-copy"}
+                    onClick={(event) => {
+                        if (stopPropagation) {
+                            event.preventDefault();
+                            event.stopPropagation();
+                        }
+                        void onCopy();
+                    }}
+                    size="small"
```
haschek, line 41: As this is a shared component it should probably not set display configuration without the option to overwrite it. `CompactCopyUriButton` could extend `IconButtonProps`, or it could offer a `iconButtonProps` property.

workspace/src/app/views/taskViews/linking/evaluation/tabView/LinkingEvaluationTabview.scss
```diff
@@ -17,6 +17,27 @@
...
+        column-gap: $eccgui-size-inline-whitespace * 0.5;
+        display: inline-flex;
+        max-width: 100%;
+    }
+
+    .diapp-linking-evaluation__selectable-tag,
+    .diapp-linking-evaluation__selectable-tag *,
+    .diapp-linking-evaluation__selectable-value-text,
+    .diapp-linking-evaluation__selectable-value-text * {
+        -webkit-user-select: text !important;
```
haschek: Is this `!important` statement necessary? If yes, then maybe the original problem should be fixed somehow else. In the storybook all tags and tag lists do not prevent selecting texts.

workspace/src/app/views/taskViews/linking/evaluation/tabView/LinkingEvaluationTabview.scss
```diff
@@ -17,6 +17,27 @@
...
+        user-select: text !important;
+    }
+
+    .diapp-linking-evaluation__selectable-value-text {
```
haschek: What is the intention here? This class is also used for text inside tags. So the text is supposed to break on multiple lines inside the tag?

workspace/src/app/views/taskViews/transform/evaluation/tabView/TransformEvaluationTabView.scss
```diff
@@ -18,4 +18,25 @@
     .ecc-silk-mapping__treenav {
         margin: 0px;
     }
+
```
haschek, line 23: Style rules are exact copies of other rules, just with different class names. Elements should share some class names that the rules need to exists (and to get maintained) only one time.

haschek (changes requested):

workspace/src/app/views/shared/CompactCopyUriButton.tsx
andreas-schultz: I've improved the CompactCopyUriButton component a little bit also covering your request.

workspace/src/app/views/taskViews/linking/evaluation/tabView/LinkingEvaluationTabview.scss
andreas-schultz: No, it is not necessary, I'll remove it.

workspace/src/app/views/taskViews/linking/evaluation/tabView/LinkingEvaluationTabview.scss
andreas-schultz: I will remove this class.

PR #1073, opened 2026-07-06
Fix gui-elements dependency mismatch by repointing the submodule to the real v26.0.0 release

edufuga:
`workspace/package.json` requires `@eccenca/gui-elements` `^26.0.0`, but the `libs/gui-elements` submodule was pinned to a `25.2.0` commit, so that requirement could never resolve. Every build on `develop` inherits this and fails at `yarn install`, including CI.

Root cause: a June 10 commit correctly upgraded to a real `26.0.0` (submodule pointer and version requirement bumped together, with matching source changes). A later "forward gui elements" commit repointed the submodule again, but landed it on an unrelated feature branch still on `25.2.0` instead of moving forward on the release line, silently reintroducing the conflict without touching the version requirement. (See 2723ff0f98366dc86c1cc2cfbdee3f5f5bf2a6d7)

This PR repoints the submodule to the current tip of `gui-elements`' `release/v26.0.0` branch, which carries a genuine final `26.0.0` release. No change to `workspace/package.json` — the requirement was correct the whole time. `yarn.lock` is regenerated to match; most of that diff is unrelated dependency re-resolution, not actual version changes.

**Verified:**
- `yarn install --frozen-lockfile` completes clean, no prompt, no warnings.
- `sbt compile` succeeds end to end on top of `develop`.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -99,6 +99,15 @@
     js-tokens "^4.0.0"
     picocolors "^1.1.1"
 
+"@babel/code-frame@^7.29.7":
+  version "7.29.7"
+  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.29.7.tgz#f2fbbfea87c44a21590ec515b778b2c26d8866e7"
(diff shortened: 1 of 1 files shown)


haschek:
Just set gui elements to `*` please.

PR #1089, opened 2026-07-20
SPARQL Update operator: outputProperties values come from the connected output task, so they do not need an input port

robertisele:
# SPARQL Update operator: outputProperties values come from the connected output task, so they do not need an input port

## Problem

The SPARQL Update operator required a connected input port even when a template only referenced `$outputProperties` values — properties of the operator's own output task configuration, not of an entity or the connected input task. Since `$outputProperties` is available regardless of whether an input is connected, this forced users to wire up a dummy input dataset that was never actually read.

## Fix

`SparqlLegacyTemplate.requiresInput` now checks each referenced variable's scope and only requires an input port if at least one variable is scoped to something other than `outputProperties` — a template referencing only `$outputProperties` now correctly reports `requiresInput = false` and runs once against an empty entity, exactly as a static template would.

A template with zero real inputs can leave a workflow node holding a trailing vacant input slot, an artifact of how the workflow editor vacates a port on disconnection. `Workflow.createNormalized` now trims trailing vacant slots from every operator's and dataset's input list on construction, while preserving interior vacant slots so the positional meaning of the remaining ports is unaffected.

## Testing

`SparqlTemplateVelocityTest` covers the outputProperties-only case end to end: `requiresInput` is false, the input schema is empty, and the template renders correctly with no entity; templates referencing `$inputProperties` or mixing an entity value with `$outputProperties` still correctly require input. `WorkflowTest` covers both new trimming shapes: a trailing-only vacant port dropped, and a leading vacant port preserved alongside an interior connection.


edufuga (approved):

edufuga:
**Validation report — CMEM-7573**

Ran an independent code review of the two commits on this branch. No defects found. Confirmed:

- `SparqlLegacyTemplate.requiresInput`'s fix is exhaustive over the real variable scopes (row / `inputProperties` / `outputProperties`) and correctly scoped to the Velocity/Simple engine only — Jinja already handled this case correctly and is untouched by the change.
- `Workflow.createNormalized`'s new trimming correctly distinguishes trailing (dropped) from interior (preserved) vacant input slots, with no analogous gap on the output side.
- Every real `Workflow` construction/persistence path — XML project storage, JSON storage/REST API, `WorkflowBuilder` — routes through `createNormalized` and receives the trim.
- Full relevant existing suites run unmodified, all passing: `SparqlTemplateVelocityTest` (9/9), `SparqlJinjaTemplateTest` (12/12), `WorkflowTest` (17/17), JSON round-trip (16/16).

Two follow-up commits closed one of two open questions from the review: added tests exercising the `outputProperties`-only exemption through Velocity's undocumented-method fallback path (not just the four recognized methods), plus a fully static template case; and added boundary-shape tests for `createNormalized`'s trimming (all-vacant → empty, no-vacant → unchanged).

Still open, not a defect: whether an interior vacant slot's positional meaning is actually consumed at workflow-execution time, or matters only for editor rendering and round-trip fidelity. Verifying that needs the executor exercised together with the editor client — outside what unit tests alone can settle.

PR #1090, opened 2026-07-21
The input source of transform and linking tasks is now optional (CMEM-4956)

robertisele:


silk-rules/src/main/scala/org/silkframework/rule/DatasetSelection.scala
```diff
@@ -48,21 +51,30 @@ case class DatasetSelection(@Param(label = "Input", value = "The input that is g
                                     "The restriction is ignored for non-RDF datasets.")
                             restriction: Restriction = Restriction.empty) extends PluginObjectParameter {
 
+  /** The configured input task, if any. */
+  def inputTaskId: Option[Identifier] = inputId.value
+
+  /** The configured input task or a clear validation error if none is configured. */
+  def requiredInputId(inputName: String = "input source"): Identifier = {
```
andreas-schultz, line 59: This is used in the EvaluateTransformApi.evaluateSpecificRule controller method. We should probably create a follow-up to handle this correctly in a workflow context, i.e. when the transform evaluation tab is opened inside a workflow.

silk-rules/src/main/scala/org/silkframework/rule/LinkSpec.scala
```diff
@@ -50,10 +50,10 @@ import scala.xml.Node
   description =
       """Generates links between instances from different sources according to a link specification."""
 )
-case class LinkSpec(@Param(label = "Source input", value = "The source input to select.")
-                    source: DatasetSelection = DatasetSelection("SourceDatasetSelection", Uri(""), Restriction.empty),
-                    @Param(label = "Target input", value = "The target input to select.")
-                    target: DatasetSelection = DatasetSelection("TargetDatasetSelection", Uri(""), Restriction.empty),
+case class LinkSpec(@Param(label = "Source input", value = "The source input to select. If left empty, the input must be provided by connecting it in a workflow.")
```
andreas-schultz: Maybe it should also be mentioned that (currently) auto-completion will be affected when left empty.

robertisele: I have added a note here and on the transform spec to make this more clear to the user.

silk-rules/src/main/scala/org/silkframework/rule/DatasetSelection.scala
robertisele: There seems to be a ticket for this already: CMEM-5398

PR #1096, opened 2026-07-28
Auto-terminate SPARQL Update queries before batching

edufuga:
# CMEM-7867: Auto-terminate SPARQL Update queries before batching

SPARQL Update queries are joined into a batch by plain string concatenation, with no check that the result is valid SPARQL. A query missing its trailing semicolon produces an invalid batch. Batching now terminates every query before joining it.

## The gap this closes

A SPARQL Update task validates its template when the task is configured. For Simple and Velocity templates without rawUnsafe output, validation renders one example query, doubles it, and checks that the result parses — a missing semicolon fails configuration before the task can be saved. Two cases skip this: rawUnsafe templates, where arbitrary SPARQL can't be checked against a synthetic example, and Jinja templates, where validation never renders and only confirms referenced variables exist. For both, a missing semicolon passes configuration unnoticed and only surfaces at execution, as a parse failure.

## Termination at the batching point

Every rendered query is trimmed of trailing whitespace and given a semicolon if it doesn't already end in one, right before it's joined into the batch — the one point every query passes through, regardless of engine, rawUnsafe setting, or batch size. Template validation and query rendering are unchanged.

## A documented limitation

A query ending in a SPARQL comment still breaks: the added semicolon lands after the comment text, where it's swallowed rather than terminating the statement. The affected code doesn't parse or strip comments; this case isn't fixed, but a test pins it down.

## Tests

Three new unit tests confirm the termination behavior. An already-terminated query passes through unchanged across two batched updates. Trailing whitespace after an existing semicolon is stripped rather than producing a double semicolon. A missing semicolon is appended after trimming.

Four new integration tests exercise the fix end to end. A Jinja template missing its semicolon is corrected across two entities. A rawUnsafe Velocity template missing its semicolon is corrected the same way. A malformed rawUnsafe query, missing both a closing brace and its semicolon, gets terminated but not otherwise fixed — the query still fails to parse, showing the change adds termination only, not validation. The comment case shows the same thing: the resulting batch still fails to parse after termination.

https://jira.eccenca.com/browse/CMEM-7867


silk-plugins/silk-plugins-rdf/src/main/scala/org/silkframework/plugins/dataset/rdf/executors/LocalSparqlUpdateExecutor.scala
```diff
@@ -118,7 +118,7 @@ case class BatchSparqlUpdateEmitter[U](f: String => U, batchSize: Int, reportUpd
     }
     queryCount += 1
 
-    sparqlUpdateQueries.append(query)
+    sparqlUpdateQueries.append(BatchSparqlUpdateEmitter.ensureTerminated(query))
```
robertisele: Wouldn't this also add a semicolon to empty queries?

edufuga: Good catch! Fixed.

PR #1105, opened 2026-08-21
MCP improvements, round 2

robertisele:


workspace/src/app/views/shared/modals/CreateArtefactModal/CreateArtefactModal.tsx
```diff
@@ -421,6 +421,15 @@ export function CreateArtefactModal() {
         [cachedArtefactProperties],
     );
 
+    /** The parameters the current plugin declares; form fields of these names are parameters, not meta data. */
```
andreas-schultz, line 424: This seems to be an incomplete workaround for an edge case. It would probably be better to model the data differently in the form instead, i.e. separate meta data values from plugin parameter values.

andreas-schultz: This will be handled in another ticket,

PR #1110, opened 2026-09-02
Add contains-mode matching to FilterByRegex (CMEM-8042)

edufuga:
# Add contains-mode matching to FilterByRegex (CMEM-8042)

## Description

FilterByRegex now matches on substring containment as well as full-value equality, through a single `contains` flag that composes cleanly with the existing `negate` inversion.

## Behavior

By default, a value is kept only when the regex matches it in full. Setting `contains` relaxes that: a value is kept if the pattern matches anywhere within it. `negate` still inverts whichever of the two decisions is in effect — the same value can be excluded on a full match, excluded on a partial match, or either, from the same two flags.

## Implementation

The match check is a single predicate, chosen by `contains`: match the whole value, or find the pattern anywhere within it. `negate` wraps that predicate rather than branching around it separately, so the same logic handles all four flag combinations without special-casing any of them.

## Tests

Coverage comes from the plugin's own worked examples rather than a single hand-written check, covering both matching modes together with negation and the sharper edges around substring matching — an anchored pattern, a zero-width pattern that matches everywhere, a regex that fails to compile. Three additional checks establish the algebraic relationships directly: negation produces the exact complement of a match set whether or not contains is also set, and contains-mode always keeps a superset of what full-match keeps for the same input.

### Missing-input crash

Connecting no input to FilterByRegex used to crash with an unguarded `NoSuchElementException`. It now throws a `ValidationException` with a message naming the problem. A connected input that carries zero values is unaffected — it already just produced an empty result.

### Documentation

FilterByRegex only had a one-line description. It now has a real documentation page, walking through the regex-escaping rules and why a missing input and an empty one behave differently.

### Related plugins

FilterByRegex cross-referenced one plugin, and that description had gone stale the moment `contains` landed — it still described full-string-only matching. It's now cross-referenced with six related transformers, each side naming the behavioral difference instead of a generic note.


silk-rules/src/main/resources/org/silkframework/rule/plugins/transformer/filter/FilterByRegex.md
```diff
@@ -0,0 +1,20 @@
+## Description
+
+The Filter by regex plugin keeps or drops values based on a regular expression. By default, a value is kept only if it matches the regex in full; setting `contains` relaxes that to keep any value the regex matches anywhere within. Setting `negate` inverts whichever of those two decisions is in effect.
+
+## Notes on regular expressions
+
+Attention: regex metacharacters in the pattern have to be escaped to be matched literally, e.g. a literal dot needs `\\.`, not `.`, which otherwise means "any character."
```
robertisele, line 7: While `\\.` is the correct way to escape `\.` in Java/Scala strings, this will confuse UI users. In the UI, users need to enter `\.`  (single slash)

silk-rules/src/main/resources/org/silkframework/rule/plugins/transformer/filter/FilterByRegex.md
```diff
@@ -0,0 +1,20 @@
...
+
+A pattern that can match an empty string, such as `a*`, matches every value under `contains`, since an empty match always "occurs somewhere" in any string.
+
+A regex that fails to compile throws a `PatternSyntaxException` rather than silently doing nothing.
```
robertisele, line 11: I think we should clarify that it only throws that during execution (I can save an operator with an invalid regex just fine)

silk-rules/src/main/resources/org/silkframework/rule/plugins/transformer/filter/FilterByRegex.md
```diff
@@ -0,0 +1,20 @@
...
+
+### Note for advanced users
+
+A compilation of the available constructs for building regular expressions is available in the
+[API of the Java `Pattern`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/regex/Pattern.html#sum).
```
robertisele, line 16: We are using JDK 25, so we might just as well update that link :)

robertisele (changes requested):

silk-rules/src/main/resources/org/silkframework/rule/plugins/transformer/filter/FilterByRegex.md
edufuga: I know, I just didn't find the API documentation for the newer Java versions.

edufuga: Updated to https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/util/regex/Pattern.html (the `#sum` part was wrong).

silk-rules/src/main/resources/org/silkframework/rule/plugins/transformer/filter/FilterByRegex.md
edufuga: Good catch. I wrote that in "programming-mode". The double-escape is indeed wrong there.

robertisele (approved):

PR #1113, opened 2026-09-11
Add configurable separator to InputHashTransformer (CMEM-8043)

edufuga:
# Add configurable separator to InputHashTransformer (CMEM-8043)

## Description

InputHashTransformer now accepts an optional separator between values, reusing the same `glue` parameter and escaping rules `ConcatTransformer` already defines.

## Behavior

The transformer still reduces any number of connected ports and values to exactly one hash value, and with no separator configured, behaves exactly as before. Once a separator is set, it treats a port boundary the same as any other adjacent pair rather than as something special — so moving a value from one port to another never changes the hash, as long as the overall value order stays the same.

## Implementation

The separator is applied once, while flattening the per-port value sequences into the single ordered sequence that gets concatenated and hashed — the hashing step itself stays untouched and has no notion of separators or ports at all. That separator is parsed using `ConcatTransformer`'s own escaping rules rather than a second implementation of the same semantics, so the two plugins can't quietly drift apart on what an escape sequence means.

## Tests

Coverage comes from two complementary sources. Transform examples cover the sharper edges directly: a value identical to the separator string itself, which hashes as ordinary content rather than being specially escaped, and a single value on its own, where the separator never applies because there's no adjacent pair to insert it between. A property-based unit test composes `ConcatTransformer` and `InputHashTransformer` independently, rather than reimplementing the hashing logic as its own reference, and verifies across several algorithms and separator values that a value split across two ports and the same value combined on one port always hash identically.

### Documentation

`InputHashTransformer.md` stated flatly that no separator was ever inserted between values — true when it was written, false the moment `glue` landed. It's been rewritten to document the parameter, and a table that judged algorithms instead of describing them ("weak," "recommended," bare comparisons) was cut down to the verifiable facts underneath those labels.

### Related plugins

InputHashTransformer cross-referenced two other transformers; it's now cross-referenced with five, each new pair naming the specific behavioral difference a user might otherwise miss rather than a generic note.

### DIFF
--- a/silk-rules/src/main/scala/org/silkframework/rule/plugins/transformer/value/InputHashTransformer.scala
+++ b/silk-rules/src/main/scala/org/silkframework/rule/plugins/transformer/value/InputHashTransformer.scala
@@ -1,18 +1,22 @@
 package org.silkframework.rule.plugins.transformer.value
 
 import org.silkframework.rule.annotations.{TransformExample, TransformExamples}
+import org.silkframework.rule.plugins.transformer.combine.ConcatMultipleValuesTransformer
+import org.silkframework.rule.plugins.transformer.combine.ConcatTransformer
+import org.silkframework.rule.plugins.transformer.combine.ZipTransformer
 import org.silkframework.rule.plugins.transformer.replace.MapTransformerWithDefaultInput
+import org.silkframework.rule.plugins.transformer.value.InputHashTransformer.SeqIntersperseOps
 import org.silkframework.runtime.plugin.annotations.{Param, Plugin, PluginReference}
 
 import java.security.MessageDigest
 
-
 /**
  * Combines all input values across all connected ports into a single hash.
  *
- * Values are fed sequentially into a single [[java.security.MessageDigest]] instance — port 1 first, then port 2,
- * and so on; within each port, values are processed in order. No separator is inserted between values or ports.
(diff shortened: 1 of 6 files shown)


silk-rules/src/main/scala/org/silkframework/rule/plugins/transformer/value/InputHashTransformer.scala
```diff
@@ -91,23 +260,43 @@ import java.security.MessageDigest
...
 
-  override def apply(values: Seq[Seq[String]]): Seq[String] = {
-    def updateAll(digest: MessageDigest): Unit =
-      values.flatten.foreach(updateWith(digest, _))
+  // glue with escaped char sequences (\\, \n, \t) converted to actual character.
+  lazy val parsedGlue: String = ConcatTransformer.parseGlue(glue)
+
+  /** Updates the digest with the UTF-8 encoding of all values. */
+  private def updateAll(digest: MessageDigest)(values: Seq[Seq[String]], separator: String): Unit =
+    values.flatten.intersperse(separator).foreach(updateWith(digest, _))
```
robertisele: It looks like this would create a new list for each value. Maybe this could be a bit more efficient by just using a plain for loop that just calls `updateWith` twice in each cycle?

robertisele (changes requested):

PR #1119, opened 2026-09-21
Fix accessibility of project task tabs

andreas-schultz:
- Replace button selectors with keyboard-accessible tabs.
- Add ARIA selection state and tab/panel relationships.
- Indicate links that open new tabs visually and through accessible descriptions.
- Add tests


workspace/src/app/views/shared/projectTaskTabView/ProjectTaskTabView.tsx
```diff
@@ -412,54 +427,79 @@ export function ProjectTaskTabView({
...
+                                    onChange={(id) => changeTab(String(id))}
+                                    allowScrollbars
+                                    tabs={inlineTabs.map((tabItem) => ({
+                                        id: tabItem.id ?? "",
+                                        title: tLabel(tabItem.label ?? ""),
+                                        "data-test-id": "taskView-" + tabItem.id,
+                                    }))}
+                                />
+                            )}
+                            <span id={externalLinkDescriptionId} hidden>
```
haschek: This is also added even if no links are available that could use this description.

haschek (changes requested):

